PR #8, opened 2025-06-23
fix: fixed the volume update logic in try_match_instantly in direct_impl.rs

### LINKED ISSUE #7: fix: The volume update logic in try_match_instantly is flawed in direct_impl.rs
### Problem: 
In both direct_impl.rs, the try_match_instantly function doesn't update the bucket states if partially filled.

### Correct Behavior: 
The partially filled order must also be reflected in the buckets state.

### Impact: 
This bug does not reduce the partially filled orders.

### Affected Files:
orderbook/src/direct_impl.rs

RajasreePonnada:
`572e03f` - fixed no volume updation after partial filling on any side
`570f62d` - test: tests the try_match_instantly that has a volume update flaw before update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of order volume tracking in the order book during partial matches, ensuring price bucket volumes are updated correctly after each trade.

* **Tests**
  * Added a new test to verify that partial matches correctly update the remaining order volume and price bucket totals.

### DIFF
--- a/orderbook/src/direct_impl.rs
+++ b/orderbook/src/direct_impl.rs
@@ -232,6 +232,16 @@ impl OrderBookDirectImpl {
                 if maker_filled {
                     self.remove_order(maker_key);
                 }
+                else {
+                    let buckets = if maker_action == OrderAction::Ask {
+                        &mut self.ask_price_buckets
+                    } else {
+                        &mut self.bid_price_buckets
+                    };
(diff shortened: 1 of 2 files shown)


esciiee (approved):
LGTM

rahulghangas (approved):

PR #10, opened 2025-06-23
fix: In direct_impl.rs, fixed to perform instant matching 

### LINKED ISSUE #5: fix: Correct move_order logic in naive and direct implementations in the Orderbook
### Problem: 
In both naive_impl.rs and direct_impl.rs, the move_order function incorrectly handles a price change. It performs a remove followed by a direct insert at the new price level.

### Correct Behavior: 
A moved order must be treated as a new incoming taker order. The correct sequence is remove -> update price -> try_match_instantly -> insert (only if not fully filled).

### Impact: 
This bug allows the creation of an invalid orderbook where the best bid price is higher than the best ask price, which should be impossible in a matching engine.

### Affected Files:
orderbook/src/naive_impl.rs
orderbook/src/direct_impl.rs

RajasreePonnada:
- The move_order function in direct_impl.rs has been rewritten to follow the correct remove -> match -> insert sequence.
- An order being moved is now correctly treated as a taker order at its new price.

## Summary by CodeRabbit

* **New Features**
  * Orders that are moved within the order book now attempt to match immediately at their new price, potentially resulting in instant trades.

* **Bug Fixes**
  * Orders that are fully matched during a move are no longer reinserted into the order book.

### DIFF
--- a/orderbook/src/direct_impl.rs
+++ b/orderbook/src/direct_impl.rs
@@ -637,7 +637,16 @@ impl<'a> OrderBook<'a> for OrderBookDirectImpl {
 
         order_to_move.price = cmd.price();
 
-        self.insert_order(order_to_move);
+        cmd.action = order_to_move.action;
+        cmd.size = order_to_move.size; 
+        
+        let total_filled = self.try_match_instantly(cmd, order_to_move.filled);
+        order_to_move.filled = total_filled;
+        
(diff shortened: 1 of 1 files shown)


rahulghangas (approved):
LGTM

rahulghangas (approved):

PR #14, opened 2025-06-24
feat: added missing state_hash and idiomatic state_hash definition

RajasreePonnada:
-  Implement the standard Rust `Hash` trait for UserProfile. The Java version has a `stateHash()` method used for data integrity checks. This is crucial for verifying that state is consistent across the system. The most robust way is to use the existing `BorshSerialize` implementation to get a byte representation and then hash it.
- `symbol_specification.rs` uses a manual `state_hash()` method. This is an inconsistency that could lead to bugs later if a field is added to the struct but not the manual hash function. Hence, made it idiomatic.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating consistent hash values for user profiles and symbol specifications, improving data integrity verification.

* **Bug Fixes**
  * Updated tests to use the new hashing approach for symbol specifications, ensuring accurate hash comparisons.

* **Chores**
  * Improved assertion message formatting in market data snapshot tests for clearer test output.

### DIFF
--- a/common/src/model/symbol_specification.rs
+++ b/common/src/model/symbol_specification.rs
@@ -1,5 +1,5 @@
 use super::enums::SymbolType;
-use borsh::{BorshDeserialize, BorshSerialize};
+use borsh::{BorshDeserialize, BorshSerialize, to_vec};
 use serde::{Deserialize, Serialize};
 use std::hash::{Hash, Hasher};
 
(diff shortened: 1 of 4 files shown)


esciiee (approved):
Incroprate changes suggested by coderabbit and it should be good.

rahulghangas (approved):

rahulghangas (approved):

PR #29, opened 2025-06-26
feat: add main.rs to Use and test All Components in Exchange Initialization and Final Event Assertion

RajasreePonnada:
- The `main.rs` is added to ensure that all core components of the exchange are properly initialized and utilized in the test pipeline. 
- The `init_exchange` function now wires together all processors and handlers, and the main test flow finally asserts that the correct event is received after a simulated client order. 
- This ensures the full pipeline is exercised, from command submission to event handling, providing a integration test for the exchange core.

## Summary by CodeRabbit

* **New Features**
  * Modular exchange entrypoint that wires core pipeline (risk, matching, journaling, events) with an async API and a sample gateway that runs a synthetic order and validates event flow.
  * Emits and validates order events through the async handler.

* **Style**
  * Standardized debug log formatting and minor import reordering for clearer logs.

* **Chores**
  * Added a new processors crate to the workspace and enabled workspace usage for a dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 # Workspace crates containing our data models and order book logic
 common = { path = "./common" }
 orderbook = { path = "./orderbook" }
+processors = { path = "./processors" }
 
 # High-level async runtime and message passing channels
 tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -17,12 +18,13 @@ tracing-subscriber = "0.3"
...
 # Trait for async functions in traits
 async-trait = "0.1"
+hashbrown.workspace = true
 
 [workspace]
 resolver = "2"
 members = [
     "common",
-    "orderbook",
+    "orderbook", "processors",
```
rahulghangas:
```suggestion
    "orderbook",
    "processors",
```

rahulghangas (changes requested):

PR #65, opened 2025-09-25
feat: add gateway core networking

esciiee:
Modified Networking Crate,
Note that the ./order directory is self generated from sbe-generator docker file in ./docker/Dockerfile

## Summary by CodeRabbit

* **New Features**
  * SBE-based OrderCommand serialization + new SBE crate and generator tooling.
  * Gateway client with handshake, dedicated channels, retries and stateful messaging.
  * Core server managing gateway sessions, ports and lifecycle.
  * Networking helpers for robust message delivery and port/session allocation.

* **Refactor**
  * Public networking API reorganized from publisher/subscriber to client/server model with new handlers.

* **Tests**
  * Added client–server integration test; removed legacy publisher/subscriber test suites.

* **Chores**
  * Workspace manifest and dependency updates; default max message size aligned to ORDERCOMMANDSIZE.


order/src/order_command_message_codec.rs
```diff
@@ -0,0 +1,332 @@
...
+            let offset = self.offset + 45;
+            self.get_buf_mut().put_u8_at(offset, value as u8)
+        }
+
+        /// REQUIRED enum
+        #[inline]
+        pub fn order_type(&mut self, value: order_type::OrderType) {
+            let offset = self.offset + 46;
+            self.get_buf_mut().put_u8_at(offset, value as u8)
+        }
```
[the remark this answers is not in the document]
esciiee: This is auto generated code, from sbe-generator, using docker/sbe

order/src/order_command_message_codec.rs
```diff
@@ -0,0 +1,332 @@
...
+            header
+        }
+
+        /// REQUIRED enum
+        #[inline]
+        pub fn command(&mut self, value: order_command_type::OrderCommandType) {
+            let offset = self.offset;
+            self.get_buf_mut().put_u8_at(offset, value as u8)
+        }
+
```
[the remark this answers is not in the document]
esciiee: This is auto generated code, from sbe-generator, using docker/sbe

rahulghangas:
@codex review

common/src/cmd/mod.rs
```diff
@@ -25,6 +32,46 @@ pub enum OrderCommandType {
...
+
+    fn try_from(val: OrderCommandType) -> Result<Self, Self::Error> {
+        match val {
+            OrderCommandType::PlaceOrder => Ok(SbeOrderCommandType::PlaceLimitOrder),
+            OrderCommandType::CancelOrder => Ok(SbeOrderCommandType::CancelOrder),
+            _ => Err(OrderCommandSerializationError::UnsupportedOrderCommandType(
+                val,
+            )),
+        }
+    }
```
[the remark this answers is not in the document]
esciiee: the reduce and move order commands are not supported as of now, therefore errors out

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+            info!("Gateway '{}': Started message polling thread", gateway_id);
+            while !shutdown.load(Ordering::SeqCst) {
+                if let Err(e) = subscription.poll(Some(&handler), 10) {
+                    error!("Gateway '{}': Error polling messages: {}", gateway_id, e);
+                    break;
+                }
+                AeronIdleStrategy::busy_spinning_idle(std::ptr::null_mut(), 0);
+            }
+            handler.release();
+        });
```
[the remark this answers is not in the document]
esciiee: THe handler passed into this is not itself leaked, but just represent a struct that implements fragment handler, and not not leaked anywhere before the thread is spawned, and here as you said, the handler guard is being moved to the spawned thread, hence there cannot be a race condition

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+
+            if result >= 0 {
+                return Ok(());
+            }
+
+            // Wait before retrying with exponential backoff
+            let delay = MESSAGE_RETRY_DELAY * (2_u32.pow(attempt as u32));
+            std::thread::sleep(delay);
+        }
+
```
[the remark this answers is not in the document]
esciiee: the max message size if confined to be the order commands message size + the header height

networking/src/lib.rs
```diff
@@ -29,29 +29,64 @@
...
-//! publisher.add_publication("aeron:ipc", ORDER_STREAM_ID)?;
+//! # VEX Networking Crate
+//!
+//! ## Overview
+//!
+//! The `networking` crate provides high-performance Aeron-based messaging for ultra-low-latency communication between the VEX Gateway and VEX Core.
+//!
+//! ## Example Usage
+//!
+//! ### Gateway (Client) Side
```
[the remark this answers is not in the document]
esciiee: this is fixed

networking/src/lib.rs
```diff
@@ -49,9 +49,8 @@
...
-pub mod client;
-pub mod publisher;
-pub mod subscriber;
+//!
 
-// Re-export main types for convenience
-pub use subscriber::{AeronSubscriber, SubscriberError};
+pub mod client;
+pub mod server;
+pub mod utils;
```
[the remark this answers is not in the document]
esciiee: updates in earlier comments

networking/src/server/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    aeron: Rc<Aeron>,
+    /// Core configuration
+    config: CoreNetworkingConfig,
+    /// Gateway state management (lock-free)
+    gateways: Rc<GatewayManager>,
+    /// Last cleanup timestamp (atomic)
+    last_cleanup_nanos: CachePadded<AtomicU64>,
+    /// shutdown flag
+    shutdown: AtomicBool,
+}
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/server/gateway_handler.rs
```diff
@@ -0,0 +1,120 @@
...
+        let address = binding.source_identity();
+
+        debug!(
+            "Gateway image available for session 0x{:x} from {}",
+            session_id, address
+        );
+
+        self.gateways
+            .set_gateway_address(session_id, address.to_string());
+    }
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/server/gateway_handler.rs
```diff
@@ -0,0 +1,120 @@
...
+        let binding = image.get_constants().unwrap();
+        let address = binding.source_identity();
+
+        debug!(
+            "Gateway image unavailable for session 0x{:x} from {}",
+            session_id, address
+        );
+
+        self.gateways.remove_gateway_address(session_id);
+    }
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/lib.rs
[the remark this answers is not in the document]
esciiee: this is intentional

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    let unavailable_logger = AeronUnavailableImageLogger {};
+    let unavailable_handler = Handler::leak(unavailable_logger);
+    aeron.add_subscription(
+        &uri,
+        stream_id,
+        Some(&available_handler),
+        Some(&unavailable_handler),
+        Duration::from_secs(1),
+    )
+}
```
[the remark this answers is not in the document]
esciiee: the implentation of handler explicitly states that the handlers are not released on drop, but are dropped when handler/release() function is called

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    let endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!("aeron:udp?endpoint={endpoint}")).unwrap();
+    aeron.add_subscription(
+        &uri,
+        stream_id,
+        Some(&Handler::leak(on_image_available)),
+        Some(&Handler::leak(on_image_unavailable)),
+        Duration::from_secs(1),
+    )
+}
```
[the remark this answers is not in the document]
esciiee: the implentation of handler explicitly states that the handlers are not released on drop, but are dropped when handler/release() function is called

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl AeronFragmentHandlerCallback for OrderCommandHandler {
+    fn handle_aeron_fragment_handler(&mut self, buffer: &[u8], _header: AeronHeader) {
+        // Deserialize OrderCommand
+        match decode_order_command(buffer) {
+            Ok(order_command) => {
+                info!(
+                    "Gateway {}: Received OrderCommand: {:?}",
+                    self.gateway_id, order_command
+                );
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/server/duologue.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn is_closed(&self) -> bool {
+        self.is_closed
+    }
+
+    pub fn close(&mut self) -> Result<(), AeronCError> {
+        self.is_closed = true;
+        self.subscription.close::<AeronNotificationLogger>(None)?;
+        self.fragment_handler.release();
+        Ok(())
+    }
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/server/gateway_manager.rs
```diff
@@ -0,0 +1,409 @@
...
+            .gateway_sessions
+            .iter()
+            .filter(|session| session.is_expired())
+            .map(|session| session.session_id)
+            .collect();
+
+        let count = expired_sessions.len();
+        for session_id in expired_sessions {
+            self.remove_gateway_session(session_id)?;
+        }
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/server/gateway_manager.rs
```diff
@@ -0,0 +1,409 @@
...
+        self.gateway_session_addresses
+            .insert(initial_session_id, gateway_address.to_string());
+
+        debug!(
+            "Allocated session 0x{:x} for gateway '{}' with ports {}, {}",
+            dedicated_session, gateway_id, ports[0], ports[1]
+        );
+
+        Ok((dedicated_session, [ports[0], ports[1]]))
+    }
```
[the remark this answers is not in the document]
esciiee: fixed

networking/src/utils.rs
```diff
@@ -0,0 +1,391 @@
...
+        while count != 0 {
+            let port = rng.gen_range(self.low..=self.high);
+            if !self.ports_used.contains(&port) {
+                result.push(port);
+                self.ports_used.insert(port);
+                count -= 1;
+            }
+        }
+
+        Ok(result)
```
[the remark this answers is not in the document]
esciiee: fixed

common/src/cmd/mod.rs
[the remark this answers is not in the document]
rahulghangas: @coderabbit add a tracking issue for this

common/src/model/enums.rs
```diff
@@ -30,6 +34,29 @@ impl OrderAction {
...
+    fn try_from(value: SbeSide) -> Result<Self, Self::Error> {
+        match value {
+            SbeSide::Bid => Ok(Self::Bid),
+            SbeSide::Ask => Ok(Self::Ask),
+            SbeSide::NullVal => Err(OrderCommandSerializationError::UnsupportedSbeOrderAction(
+                value as u8,
+            )),
+        }
+    }
+}
```
rahulghangas, line 49: Why do we need to have separate types for `SbeSide` and `Self::Side`? This seems like it is just redundant conversion

common/src/model/enums.rs
```diff
@@ -55,6 +82,35 @@ pub enum OrderType {
...
+            SbeOrderType::Ioc => Ok(Self::Ioc),
+            SbeOrderType::IocBudget => Ok(Self::IocBudget),
+            SbeOrderType::Fok => Ok(Self::Fok),
+            SbeOrderType::FokBudget => Ok(Self::FokBudget),
+            SbeOrderType::NullVal => Err(OrderCommandSerializationError::UnsupportedSbeOrderType(
+                value as u8,
+            )),
+        }
+    }
+}
```
rahulghangas, line 100: Same as above

common/src/model/enums.rs
```diff
@@ -55,6 +82,35 @@ pub enum OrderType {
...
+        match val {
+            OrderType::Gtc => SbeOrderType::Gtc,
+            OrderType::Ioc => SbeOrderType::Ioc,
+            OrderType::IocBudget => SbeOrderType::IocBudget,
+            OrderType::Fok => SbeOrderType::Fok,
+            OrderType::FokBudget => SbeOrderType::FokBudget,
+        }
+    }
+}
+
```
rahulghangas, line 113: Same as above

docker/sbe/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+WORKDIR /app
+
+# Copy the generated code
+COPY --from=builder /app/generated /app/generated
+
+# Copy the generated SBE files to a volume for extraction
+VOLUME ["/output"]
+
+# Copy generated files to output volume
+CMD ["sh", "-c", "cp -r /app/generated/* /output/ 2>/dev/null || echo 'No files to copy'; ls -la /output/; echo 'SBE Rust code generation completed'"]
```
rahulghangas: This has silent error repression. Better to use something like
```bash
CMD ["sh", "-c", "cp -r /app/generated/* /output/ && echo 'Success' || exit 1"]
```

docker/sbe/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN mkdir /app/generated && java --add-opens java.base/jdk.internal.misc=ALL-UNNAMED \
+  -Dsbe.output.dir=/app/generated \
+  -Dsbe.target.language=Rust \
+  -Dsbe.target.namespace=sbe \
+  -cp "sbe-tool/build/libs/sbe-tool-1.36.0-SNAPSHOT.jar:sbe-all/build/libs/sbe-all-1.36.0-SNAPSHOT.jar" \
+  uk.co.real_logic.sbe.SbeTool \
+  /app/resources/ordercmd.xml
+
+# Runtime stage
+FROM rust:1.87-slim
```
rahulghangas: Your runtime stage uses rust:1.87-slim just to copy files around. Since you're only extracting generated code (not compiling Rust)
Can't you use a smaller one - alpine slim?

networking/src/client/mod.rs
rahulghangas: If polling errors occur, the loop breaks without calling handler.release()
Thread panics would prevent cleanup
No join handle stored = can't wait for thread termination

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        // Set shutdown flag
+        self.shutdown.store(true, Ordering::SeqCst);
+
+        info!(
+            "VEX Gateway '{}' shut down successfully",
+            self.config.gateway_id
+        );
+        Ok(())
+    }
```
rahulghangas: shutdown doesn't actually wait

could do something like
```rust
pub struct VexGateway {
    // Add this field:
    polling_thread: Option<std::thread::JoinHandle<()>>,
    // ...
}

pub async fn shutdown(&mut self) -> Result<(), GatewayError> {
    self.shutdown.store(true, Ordering::SeqCst);
    
    // Wait for thread to finish
    if let Some(handle) = self.polling_thread.take() {
        handle.join().map_err(|_| 
            GatewayError::ProtocolError("Thread panic".into()))?;
    }
    
    Ok(())
}
```

networking/src/client/mod.rs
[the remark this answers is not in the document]
rahulghangas: @esciiee I don't think this is good to do, because changing the max message size can lea to unexpected behavior in the future

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+            "Gateway '{}': Sending OrderCommand: {:?}",
+            self.config.gateway_id, order_command
+        );
+
+        // // Calculate actual encoded size (you may need to adjust this based on your encoding)
+        // let encoded_size = std::cmp::min(buffer.len(), self.config.max_message_size);
+
+        // Send using the buffer directly
+        for attempt in 0..MESSAGE_RETRY_COUNT {
+            let result = publication.offer::<AeronReservedValueSupplierLogger>(&buffer, None);
```
rahulghangas, line 610: Send encoded length only. It is not expensive to create a slice

networking/src/client/mod.rs
rahulghangas: Why does shutdown need to be async if there is no asynchronous behavior inside

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        let mut handler = Handler::leak(fragment_handler);
+
+        let start = Instant::now();
+        while start.elapsed() < CONNECT_TIMEOUT {
+            subscription.poll(Some(&handler), 10)?;
+            if let Some(response) = shared_response.lock().unwrap().take() {
+                handler.release();
+                return Ok(response);
+            }
+            // Sleeping breifly here. Larfer sleep as latency is not critical during handshake
```
rahulghangas:
```suggestion
            // Sleeping breifly here. Larger sleep as latency is not critical during handshake
```

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+            std::thread::sleep(Duration::from_millis(10));
```
rahulghangas, line 459: Can you justify the above comment and why this sleep is needed?

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+            state: Arc::new(RwLock::new(GatewayState::Disconnected)),
+            session_id: None,
+            encryption_key: None,
+            core_publication: None,
+            shutdown: Arc::new(AtomicBool::new(false)),
+        })
+    }
+
+    /// Starts the gateway and establishes connection to VEX Core
+    pub fn start<AeronFragmentHandlerHandlerImpl>(
```
rahulghangas: `AeronFragmentHandlerHandlerImpl` why does this have the word `Handler` twice in it

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        }
+
+        let publication = self
+            .core_publication
+            .as_ref()
+            .ok_or(GatewayError::VexCoreNotStarted)?;
+
+        // Serialize OrderCommand
+        let mut buffer = vec![0u8; self.config.max_message_size];
+        encode_order_command(order_command.clone(), &mut buffer).map_err(|e| {
```
rahulghangas: This clone shouldn't be required

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+        std::thread::spawn(move || {
+            let mut handler = Handler::leak(handler);
+
+            info!("Gateway '{}': Started message polling thread", gateway_id);
+            while !shutdown.load(Ordering::SeqCst) {
+                if let Err(e) = subscription.poll(Some(&handler), 10) {
+                    error!("Gateway '{}': Error polling messages: {}", gateway_id, e);
+                    break;
+                }
+                AeronIdleStrategy::busy_spinning_idle(std::ptr::null_mut(), 0);
```
rahulghangas, line 428: Shouldn't we use a better idle strategy than to pin the cpu at 100% usage?

networking/src/server/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+
+            // Poll all active gateway sessions (lock-free)
+            if let Err(e) = self.gateways.poll() {
+                error!("Error polling gateways: {}", e);
+            }
+
+            // Perform periodic cleanup
+            self.periodic_cleanup()?;
+
+            AeronIdleStrategy::busy_spinning_idle(std::ptr::null_mut(), 0);
```
rahulghangas, line 158: Same as above

networking/src/server/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+        handler.release();
```
rahulghangas, line 160: This handler should be called even when the media driver causes a panic or the caller directly panics

networking/src/client/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+                parts[3].parse::<u16>(),
+                parts[4].parse::<u16>(),
+                parts[5].parse::<i32>(),
+            ) {
+                (Ok(port), Ok(control_port), Ok(encrypted_session)) => CoreResponse::Accept {
+                    dedicated_port: port,
+                    dedicated_control_port: control_port,
+                    encrypted_session,
+                    gateway_id: gateway_id.to_string(),
+                },
```
rahulghangas, line 116: `port`, `control_port`, `encrypted_session` all should have validation checks

For example
```rust
if port == 0 || control_port == 0 {
    return CoreResponse::Ignore;
}
```

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+
+                self.sender.send(order_command).unwrap_or_else(|e| {
+                    error!(
+                        "Gateway {}: Failed to send OrderCommand to channel: {:?}",
+                        self.gateway_id, e
+                    );
+                });
```
rahulghangas, line 26: Silent channel disconnection. If the receiver is dropped, send() will fail on every single message but the handler keeps running, logging errors infinitely.
```suggestion
                if let Err(e) = self.sender.send(order_command) {
                    error!(
                        "Gateway {}: Channel disconnected: {:?}. Stopping handler.",
                        self.gateway_id, e
                    );
                    self.is_channel_alive.store(false, Ordering::SeqCst);
                    // Signal upstream that handler is dead
                    return;
                }
```

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub struct OrderCommandHandler {
+    gateway_id: String,
+    sender: Sender<OrderCommand>,
+}
+
+impl AeronFragmentHandlerCallback for OrderCommandHandler {
+    fn handle_aeron_fragment_handler(&mut self, buffer: &[u8], _header: AeronHeader) {
+        // Deserialize OrderCommand
+        match decode_order_command(buffer) {
```
rahulghangas, line 14: there should be buffer validation here, what if buffer is empty or huge?

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+use tracing::{debug, error};
+
+pub struct OrderCommandHandler {
+    gateway_id: String,
+    sender: Sender<OrderCommand>,
+}
+
+impl AeronFragmentHandlerCallback for OrderCommandHandler {
+    fn handle_aeron_fragment_handler(&mut self, buffer: &[u8], _header: AeronHeader) {
+        // Deserialize OrderCommand
```
rahulghangas, line 13: Header might contain important metadata, are we sure we just want to drop it?

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+use rusteron_client::{AeronFragmentHandlerCallback, AeronHeader};
+use std::sync::mpsc::Sender;
+use tracing::{debug, error};
+
+pub struct OrderCommandHandler {
+    gateway_id: String,
+    sender: Sender<OrderCommand>,
+}
+
+impl AeronFragmentHandlerCallback for OrderCommandHandler {
```
rahulghangas, line 11: No Error Statistics/Metrics

The handler silently logs errors but doesn't track:
- Number of decode failures
- Number of send failures
- Success rate
- Processing latency

rahulghangas: @coderabbit create a tracking issue for this with a recommended implementation

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn handle_aeron_fragment_handler(&mut self, buffer: &[u8], _header: AeronHeader) {
+        // Deserialize OrderCommand
+        match decode_order_command(buffer) {
+            Ok(order_command) => {
+                debug!(
+                    "Gateway {}: Received OrderCommand: {:?}",
+                    self.gateway_id, order_command
+                );
+
+                self.sender.send(order_command).unwrap_or_else(|e| {
```
rahulghangas, line 21: How are we handling backpressure here? When full, send() blocks. But this is in a callback - blocking the entire Aeron polling thread

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+                    error!(
+                        "Gateway {}: Failed to send OrderCommand to channel: {:?}",
+                        self.gateway_id, e
+                    );
+                });
+            }
+            Err(e) => {
+                error!(
+                    "Gateway {}: Failed to decode OrderCommand: {:?}",
+                    self.gateway_id, e
```
rahulghangas: String formatting on every call
Which means for hot path, this slows down the entire exchange quite a bit. Use references or structured logging:
```
error!(
    gateway_id = %self.gateway_id,
    error = ?e,
    "Failed to decode OrderCommand"
);
```

networking/src/client/cmd_handler.rs
```diff
@@ -0,0 +1,42 @@
...
+
+                self.sender.send(order_command).unwrap_or_else(|e| {
+                    error!(
+                        "Gateway {}: Failed to send OrderCommand to channel: {:?}",
+                        self.gateway_id, e
+                    );
+                });
+            }
+            Err(e) => {
+                error!(
```
rahulghangas, line 29: Error context is weak

Add more context
```
Err(e) => {
    error!(
        "Gateway {}: Failed to decode OrderCommand: {:?}. Buffer length: {}, First bytes: {:?}",
        self.gateway_id, 
        e,
        buffer.len(),
        &buffer[..buffer.len().min(16)]  // First 16 bytes for debugging
    );
}
```

networking/src/server/cmd_handler.rs
```diff
@@ -0,0 +1,78 @@
...
+                    session_id, self.gateway_id, order_command
+                );
+
+                // Process the order command (placeholder function)
+                self.producer.publish(|cmd| {
+                    *cmd = order_command.clone();
+                });
+
+                // Serialize and send back the processed command
+                let mut response_buffer = vec![0u8; 67];
```
rahulghangas: Where is this magic number coming from? It should not be hardcoded here

networking/src/server/cmd_handler.rs
```diff
@@ -0,0 +1,78 @@
...
+        match decode_order_command(buffer) {
+            Ok(order_command) => {
+                debug!(
+                    "[{}] Gateway '{}': Received OrderCommand: {:?}",
+                    session_id, self.gateway_id, order_command
+                );
+
+                // Process the order command (placeholder function)
+                self.producer.publish(|cmd| {
+                    *cmd = order_command.clone();
```
rahulghangas, line 38: why is there a clone needed here, in the hot path?
```rust
// Store order before encode
let order_for_response = order_command.clone();

// Publish without clone
self.producer.publish(|cmd| {
    *cmd = order_command;  // Move, no clone
});

// Use cloned version for response
encode_order_command(order_for_response, &mut response_buffer)
```

networking/src/server/duologue.rs
```diff
@@ -0,0 +1,180 @@
...
+            fragment_handler: Handler::leak(fragment_handler),
+            gateway_id: gateway_id.to_string(),
+            port_data,
+            port_control,
+            is_closed: false,
+            expire_time,
+            session_id,
+            subscription,
+        })
+    }
```
rahulghangas, line 91:
```rust
pub fn new(...) -> Result<Self, AeronCError> {
    // ...
    let fragment_handler = FragmentHandler { /* ... */ };
    
    Ok(Self {
        fragment_handler: Handler::leak(fragment_handler),  // leak
        // ...
    })
}

pub fn close(&mut self) -> Result<(), AeronCError> {
    self.fragment_handler.release();  // Only released if close() is called
    // ...
}
```

Add a `Drop` impl
```rust
impl Drop for Duologue {
    fn drop(&mut self) {
        if !self.is_closed {
            if let Err(e) = self.close() {
                error!("Failed to close Duologue during drop: {:?}", e);
            }
        }
    }
}
```

networking/src/server/gateway_handler.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::rc::Rc;
+use tracing::{debug, error};
+
+use super::gateway_manager::GatewayManager;
+
+/// Handles initial handshake messages from gateways
+pub struct HandshakeMessageHandler {
+    gateways: Rc<GatewayManager>,
+    publication: AeronPublication,
+}
```
rahulghangas, line 14: This is not thread safe, especially access to `gateways`
```rust
use std::sync::Arc;

pub struct HandshakeMessageHandler {
    gateways: Arc<GatewayManager>,  // Thread-safe
    publication: AeronPublication,
}
```

networking/src/server/gateway_handler.rs
```diff
@@ -0,0 +1,131 @@
...
+                "Error processing handshake from session 0x{:x}: {}",
+                session_id, e
+            );
+        }
+    }
+}
+
+/// Handles gateway image availability events
+pub struct GatewayImageAvailableHandler {
+    gateways: Rc<GatewayManager>,
```
rahulghangas: This is not thread safe, use Arc

networking/src/server/gateway_handler.rs
```diff
@@ -0,0 +1,131 @@
...
+        );
+
+        self.gateways
+            .set_gateway_address(session_id, address.to_string());
+    }
+}
+
+/// Handles gateway image unavailability events
+pub struct GatewayImageUnavailableHandler {
+    gateways: Rc<GatewayManager>,
```
rahulghangas: This is not thread safe, use Arc

networking/src/server/cmd_handler.rs
```diff
@@ -0,0 +1,78 @@
...
+        // Deserialize OrderCommand
+        match decode_order_command(buffer) {
+            Ok(order_command) => {
+                debug!(
+                    "[{}] Gateway '{}': Received OrderCommand: {:?}",
+                    session_id, self.gateway_id, order_command
+                );
+
+                // Process the order command (placeholder function)
+                self.producer.publish(|cmd| {
```
rahulghangas: No backpressure handling

networking/src/server/duologue.rs
```diff
@@ -0,0 +1,180 @@
...
+        aeron: &Aeron,
+        local: &str,
+        gateway_id: &str,
+        owner: &str,
+        port_data: u16,
+        port_control: u16,
+        session_id: i32,
+        producer: MultiProducer<OrderCommand, MultiConsumerBarrier>,
+    ) -> Result<Self, AeronCError> {
+        let expire_time = (SystemTime::now() + Duration::from_secs(1_000_000))
```
rahulghangas: Magic number here

networking/src/server/duologue.rs
```diff
@@ -0,0 +1,180 @@
...
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+
+        let publication = new_publication_with_mdc_and_session(
+            aeron,
+            local,
+            port_control,
```
rahulghangas, line 51: No validation for port (not to be 0)

networking/src/server/duologue.rs
```diff
@@ -0,0 +1,180 @@
...
+};
+use tracing::{error, info};
+
+pub const DUOLOGUE_STREAM_ID: i32 = 1002;
+
+pub struct Duologue {
+    pub fragment_handler: Handler<FragmentHandler>,
+    pub session_id: i32,
+    pub gateway_id: String,
+    pub subscription: AeronSubscription,
```
rahulghangas: `fragment_handler` `subscription` should be private

networking/src/server/gateway_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+/// session allocation, resource management, and cleanup operations.
+pub struct GatewayManager {
+    /// Active gateway sessions mapped by session ID
+    gateway_sessions: DashMap<i32, Duologue>,
+    /// Gateway addresses mapped by session ID
+    gateway_session_addresses: DashMap<i32, String>,
+    /// Connection count per address for rate limiting
+    address_connection_count: DashMap<String, AtomicU64>,
+    /// Aeron messaging instance
+    aeron: Rc<Aeron>,
```
rahulghangas: This is not thread safe

networking/src/server/gateway_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+            .insert(initial_session_id, gateway_session);
+        self.gateway_session_addresses
+            .insert(initial_session_id, gateway_address.to_string());
+
+        debug!(
+            "Allocated session 0x{:x} for gateway '{}' with ports {}, {}",
+            dedicated_session, gateway_id, ports[0], ports[1]
+        );
+        Ok((dedicated_session, [ports[0], ports[1]]))
+    }
```
rahulghangas, line 389: Incomplete cleanup on error path
```rust
fn allocate_gateway_session(...) -> Result<(i32, [u16; 2]), ServerError> {
    // Update connection count
    counter.fetch_add(1, Ordering::Relaxed);
    
    let ports = match self.port_allocator.allocate(2) {
        Ok(p) => p,
        Err(e) => {
            counter.fetch_sub(1, Ordering::Relaxed);  // This is fine
            return Err(...);
        }
    };
    
    let dedicated_session = match self.session_allocator.allocate() {
        Ok(s) => s,
        Err(e) => {
            self.port_allocator.free(ports[0]);
            self.port_allocator.free(ports[1]);
            counter.fetch_sub(1, Ordering::Relaxed);  // This is fine
            return Err(...);
        }
    };
    
    let gateway_session = match Duologue::new(...) {
        Ok(session) => session,
        Err(e) => {
            // ✓ Cleanup looks good
            self.port_allocator.free(ports[0]);
            self.port_allocator.free(ports[1]);
            self.session_allocator.free(dedicated_session);
            counter.fetch_sub(1, Ordering::Relaxed);
            return Err(...);
        }
    };
    
    // Store session
    self.gateway_sessions.insert(initial_session_id, gateway_session);
    self.gateway_session_addresses.insert(initial_session_id, gateway_address.to_string());
    // What if this insert fails (out of memory)? gateway_session leaked!
```

rahulghangas: Use RAII
```rust
struct AllocationGuard<'a> {
    manager: &'a GatewayManager,
    ports: Option<[u16; 2]>,
    session: Option<i32>,
    counter: Option<(String, &'a AtomicU64)>,
}

impl Drop for AllocationGuard<'_> {
    fn drop(&mut self) {
        if let Some([p1, p2]) = self.ports {
            self.manager.port_allocator.free(p1);
            self.manager.port_allocator.free(p2);
        }
        if let Some(s) = self.session {
            self.manager.session_allocator.free(s);
        }
        if let Some((_, counter)) = &self.counter {
            counter.fetch_sub(1, Ordering::Relaxed);
        }
    }
}

impl AllocationGuard<'_> {
    fn commit(mut self) {
        self.ports = None;
        self.session = None;
        self.counter = None;
    }
}
```

networking/src/utils.rs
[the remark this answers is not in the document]
rahulghangas: @esciiee this is indeed a massive memory leak

rahulghangas, line 168: Every call to these subscription functions leaks 2 handlers. With hundreds or thousands of subscriptions over time, this is catastrophic.
Handlers are passed to add_subscription, but there's no way to retrieve them later to call release(). The subscription doesn't own them.

Either you could
- Don't leak, let Aeron manage lifetime:
- Return handlers so caller can manage:

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    aeron.add_subscription(
+        &uri,
+        stream_id,
+        Some(&available_handler),
+        Some(&unavailable_handler),
+        Duration::from_secs(1),
+    )
+}
+
+pub fn new_subscription_with_mdc_and_session(
```
rahulghangas, line 96: memory leak same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    aeron.add_subscription(
+        &uri,
+        stream_id,
+        Some(&available_handler),
+        Some(&unavailable_handler),
+        Duration::from_secs(1),
+    )
+}
+
+pub fn new_subsciption_with_handlers_and_session<
```
rahulghangas, line 121: memory leak, same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    aeron.add_subscription(
+        &uri,
+        stream_id,
+        Some(&Handler::leak(on_image_available)),
+        Some(&Handler::leak(on_image_unavailable)),
+        Duration::from_secs(1),
+    )
+}
+
+pub fn new_subscription_with_handlers<
```
rahulghangas, line 147: memory leaked, same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+>(
+    aeron: &Aeron,
+    address: &str,
+    port: u16,
+    stream_id: i32,
+    on_image_available: X,
+    on_image_unavailable: Y,
+) -> Result<AeronSubscription, AeronCError> {
+    let endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!("aeron:udp?endpoint={endpoint}")).unwrap();
```
rahulghangas: this should not be an unwrap

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+const MESSAGE_RETRY_COUNT: usize = 5;
+
+pub fn new_publication(
+    aeron: &Aeron,
+    address: &str,
+    port: u16,
+    stream_id: i32,
+) -> Result<AeronPublication, AeronCError> {
+    let endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!("aeron:udp?endpoint={endpoint}")).unwrap();
```
rahulghangas: should not be an unwrap

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    address: &str,
+    port: u16,
+    stream_id: i32,
+    session_id: i32,
+) -> Result<AeronPublication, AeronCError> {
+    let control_endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?control={control_endpoint}|control-mode=dynamic|session-id={session_id}"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    aeron: &Aeron,
+    address: &str,
+    port: u16,
+    stream_id: i32,
+) -> Result<AeronPublication, AeronCError> {
+    let control_endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?control={control_endpoint}|control-mode=dynamic"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    address: &str,
+    port: u16,
+    stream_id: i32,
+    session_id: i32,
+) -> Result<AeronPublication, AeronCError> {
+    let endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?endpoint={endpoint}|session-id={session_id}"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    aeron: &Aeron,
+    address: &str,
+    port: u16,
+    stream_id: i32,
+) -> Result<AeronSubscription, AeronCError> {
+    let control_endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?control={control_endpoint}|control-mode=dynamic"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    address: &str,
+    port: u16,
+    stream_id: i32,
+    session_id: i32,
+) -> Result<AeronSubscription, AeronCError> {
+    let control_endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?control={control_endpoint}|control-mode=dynamic|session-id={session_id}"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+    stream_id: i32,
+    session_id: i32,
+    on_image_available: X,
+    on_image_unavailable: Y,
+) -> Result<AeronSubscription, AeronCError> {
+    let endpoint = format!("{address}:{port}");
+    let uri = CString::new(format!(
+        "aeron:udp?endpoint={endpoint}|session-id={session_id}"
+    ))
+    .unwrap();
```
rahulghangas: Same as before

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+            let port = rng.gen_range(self.low..=self.high);
+            if !self.ports_used.contains(&port) {
+                result.push(port);
+                self.ports_used.insert(port);
+                count -= 1;
+            }
+        }
+
+        Ok(result)
+    }
```
rahulghangas, line 310: Time of check, Time of use bug

```rust
pub fn allocate(&self, mut count: usize) -> Result<Vec<u16>, ServerError> {
    // Check available
    if count > total.saturating_sub(used) {  // TOCTOU bug
        return Err(...);
    }
    
    // ... time passes, other threads could allocate ...
    
    while count != 0 {
        let port = rng.gen_range(self.low..=self.high);
        if !self.ports_used.contains(&port) {  // Check
            result.push(port);
            self.ports_used.insert(port);  // Use - race window!
            count -= 1;
        }
    }
```

use dashset's atomic entry api
```rust
pub fn allocate(&self, mut count: usize) -> Result<Vec<u16>, ServerError> {
    if count == 0 {
        return Ok(Vec::new());
    }
    
    let total = self.total_ports();
    let mut result = Vec::with_capacity(count);
    let mut rng = rand::thread_rng();
    let mut attempts = 0;
    let max_attempts = count * 10; // Prevent infinite loop
    
    while result.len() < count {
        if attempts >= max_attempts {
            // Rollback allocated ports
            for port in result {
                self.ports_used.remove(&port);
            }
            return Err(ServerError::ResourceAllocationError(
                format!("Could not allocate {} ports after {} attempts", count, max_attempts)
            ));
        }
        
        attempts += 1;
        
        if self.ports_used.len() >= total {
            // Rollback
            for port in result {
                self.ports_used.remove(&port);
            }
            return Err(ServerError::ResourceAllocationError(
                "No ports available".to_string()
            ));
        }
        
        let port = rng.gen_range(self.low..=self.high);
        
        // Atomic insert - returns true if actually inserted
        if self.ports_used.insert(port) {
            result.push(port);
        }
    }
    
    Ok(result)
}
```

networking/src/utils.rs
```diff
@@ -0,0 +1,411 @@
...
+                "Failed to send message: {}",
+                AeronCError::from_code(result as i32)
+            );
+        }
+        error!(
+            "Failed to send message (attempt {} of {}): {}",
+            i + 1,
+            MESSAGE_RETRY_COUNT,
+            AeronCError::from_code(result as i32)
+        );
```
rahulghangas: Double entry logging

rahulghangas (changes requested):

common/src/model/enums.rs
esciiee: sbeside comes from auto generated code by sbe, to keep two separate sides is to keep the common lib consistent and self contained, so that importing common lib imports all types

common/src/model/enums.rs
esciiee: same for this

common/src/model/enums.rs
esciiee: same for this

networking/src/server/gateway_manager.rs
esciiee: aeron is'nt shared b/w any threads

networking/src/server/gateway_handler.rs
esciiee: Gateway manager is'nt shared b/w any threads

networking/src/server/gateway_handler.rs
esciiee: we dont need to use arc, unless they are shared b/w multiple threads

networking/src/server/gateway_handler.rs
esciiee: same

networking/src/client/cmd_handler.rs
esciiee: contains frame_id, initial frame id, etc..
needed when the received message is too large to fit in a single frame, which equals MTU, way higher than out usage

networking/src/client/cmd_handler.rs
esciiee: we dont do any allocations in the decode fn, if the buffer is invalid, the decode function will return err

networking/src/client/cmd_handler.rs
esciiee: 1. there's no back pressure here, the gateways order pipeline is single threaded and will wait for the response from the core server before processing future orders, so at any time there's at most 1 order command.
2. send function is non-blocking, unless the receiver is dropped we can be certainly sure that the consumer will receive the command

networking/src/client/cmd_handler.rs
esciiee: the receiver should ideally never be dropped, it must be part of the struct thats does handling the order requests

networking/src/client/cmd_handler.rs
esciiee: the error contains the additional context having what part of buffer caused the error

networking/src/client/mod.rs
esciiee: port and control port's are explicit to vex-core's system, the session and additional checks are validated

networking/src/client/mod.rs
esciiee: no, on polling error the loop breaks and release is called

networking/src/client/mod.rs
esciiee: handshake is not on the hot path, and we give large sleep latency to wait for core to get the response

networking/src/client/mod.rs
esciiee: fixed

networking/src/client/mod.rs
esciiee: fixed

networking/src/server/duologue.rs
esciiee: fixed

networking/src/client/mod.rs
esciiee: fixed

networking/src/server/duologue.rs
esciiee: already validated in port allocator

networking/src/server/duologue.rs
esciiee: fixed

docker/sbe/Dockerfile
esciiee: fixed

docker/sbe/Dockerfile
esciiee: fixed

networking/src/client/mod.rs
esciiee: fixed

networking/src/client/mod.rs
esciiee: fixed

networking/src/client/mod.rs
esciiee: this is on the hotpath, waits for response from core, before processing next orders

networking/src/server/mod.rs
esciiee: same for this, the networking core responsibility is to listen for new order commands, so busy spinning

networking/src/client/cmd_handler.rs
esciiee: fixed

networking/src/server/cmd_handler.rs
esciiee: currently this pr does not include response handling, this will be removed

networking/src/server/cmd_handler.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

docker/sbe/Dockerfile
esciiee: fixed

networking/src/client/mod.rs
esciiee: we do send only the encoded length size

networking/src/server/mod.rs
esciiee: is the mdia driver panics, the client returns an error, which will result out in release eventually

networking/src/server/cmd_handler.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/server/duologue.rs
esciiee: fixed

networking/src/server/gateway_manager.rs
esciiee: rust aborts when we are out of memory, if the insertion fails then we're already doomed. apart from that the clean up here is clean and inline makes the code easier to follow without the need of introducing extra abstractions

networking/src/utils.rs
esciiee: the ports are allocated and free'd in a single main listener thread, toctau is not possible

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

networking/src/utils.rs
esciiee: fixed

rahulghangas (approved):

Issue #138, opened 2026-08-07
fix: zero-size orders are accepted, rest for free, and emit phantom trades


rahulghangas:
## Summary

No `size > 0` check exists anywhere. A zero-size order occupies a book slot and a price level, and
produces zero-size `MatcherTradeEvent`s that reach the `trades` topic.

## Reproduction

```
size=0 GTC BID -> status Placed size 0
bid levels (price, volume, orders): [(100, 0, 1)]

incoming BID @100 x5 against a resting zero-size ask -> 1 trade event, size 0
```

## Evidence

`orderbook/src/lib.rs:242`, `processors/src/risk_engine.rs:304`.

## Cross-reference

`vex-gateway` does not check `volume > 0` either.

## Suggested fix

Reject `size == 0` at ingress and in R1.

---
*Filed from a full-workspace review of `vex-core` at `2852560`, cross-referenced against `vex-gateway`. Line references validated against the tree at time of filing.*

rahulghangas:
Addressed by the fixes already open for the same defect; closing rather than opening a fourth PR touching the same guards.

**Engine side — #111 / PR #146.** `RiskEngine::reserve_funds_for_order` now rejects a zero-size order outright:

```rust
if cmd.size == 0 {
    return Err(RiskEngineError::InvalidArguments { price: cmd.price, size: cmd.size });
}
```

R1 runs before matching, so the command is marked `Rejected` and never reaches `add_to_book`. It cannot occupy a book slot, cannot hold a price level open, and cannot produce a zero-size `MatcherTradeEvent`. That is the whole impact described here. The guard is on the place-order path only, so cancels — which legitimately carry `size = 0` — are unaffected.

**Gateway side — trade-vex/vex-gateway#48 / PR trade-vex/vex-gateway#81.** `validate_order_inputs` rejects `volume == 0` at the API boundary before an `OrderCommand` is ever built, which is the "reject at ingress" half of the suggested fix and closes the cross-reference noted above.

Neither guard introduced a new error type or a new configuration value — both reuse the existing rejection paths.

Closing as addressed. Please reopen if a zero-size order is observed reaching the book once those two PRs are merged.

PR #146, opened 2026-08-14
fix: reject orders whose collateral requirement rounds to zero

rahulghangas:
## The problem

`calculate_quote_cost` is an integer **floor** division with a narrowing cast, so a small enough
notional produces a collateral requirement of exactly zero:

```rust
(numerator / denominator) as u64          // common/src/market_specification.rs:44
```

`lock_funds` succeeds when `amount == 0` even on an account with no balance, so Risk R1 accepted the
order, and settlement then subtracted 0 from locked while crediting the full `size` in base.

With the ETH/USDT scaling shipped in `vex-config` the divisor is `1e13`, so any order where
`price * size < 1e13` was free. Measured against the real engine: an account holding **zero** quote
placed a bid and received 1,000,000,000 base units.

The same cast was also the second path in: a numerator above `u64::MAX` wrapped to a small value, so
an arbitrarily large order could also be made cheap.

## The fix

Three changes, all in the reservation path:

- `calculate_quote_cost` now saturates instead of wrapping (`u64::try_from(..).unwrap_or(u64::MAX)`),
  so an oversized notional becomes unaffordable rather than cheap.
- A BID whose computed quote cost is `0` is rejected via the existing
  `RiskEngineError::InvalidArguments`, which `pre_process_command` already maps to
  `Status::Rejected`.
- `size == 0` is rejected for both sides, since a zero-size order is never legitimate and previously
  rested for free.

No new configuration, no minimum-notional knob, no new error variant — the existing error and
rejection path carry it.

## Cross-reference

`vex-gateway` is the upstream that this engine's docs assume validates notional, and it does not —
its only order validation is `market_id <= i32::MAX`. Tracked there as trade-vex/vex-gateway#48.
This engine-side guard is the backstop that holds regardless of what upstream sends.

This also closes the reachable path for #138 (zero-size orders), since a zero-size order can no
longer pass R1 to reach the book.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p processors -p common`
passes with three new tests covering a zero-notional bid, a zero-size order, and a normal bid still
locking the expected amount.

Closes #111

## Summary by CodeRabbit

* **Bug Fixes**
  * Safely handle quote cost calculations that exceed the supported value range.
  * Treat zero-priced asks as market sells when reserving collateral.
  * Reject limit asks when calculated proceeds round to zero.
  * Continue rejecting invalid or zero-size orders before funds are locked.
* **Tests**
  * Added coverage for market-sell collateral reservations and standard limit-ask reservations.

### DIFF
--- a/common/src/market_specification.rs
+++ b/common/src/market_specification.rs
@@ -41,7 +41,7 @@ impl CoreMarketSpecification {
         if denominator == 0 {
             return 0;
         }
-        (numerator / denominator) as u64
+        u64::try_from(numerator / denominator).unwrap_or(u64::MAX)
     }
 }
 
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -306,14 +306,38 @@ impl RiskEngine {
         cmd: &mut OrderCommand,
         price_cache: Arc<PriceCache>,
     ) -> Result<()> {
+        if cmd.size == 0 {
+            return Err(RiskEngineError::InvalidArguments {
+                price: cmd.price,
+                size: cmd.size,
+            });
+        }
+
         let (asset_to_lock, amount_to_lock) = match cmd.side {
             // For a Bid (buy), we lock the quote currency. Amount = price * size.
             Side::Bid => {
                 let amount = self.bid_amount(cmd, price_cache)?;
+                if amount == 0 {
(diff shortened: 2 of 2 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #147, opened 2026-08-14
fix: validate length and SBE header before decoding an order command

rahulghangas:
## The problem

`decode_order_command` was called directly on raw Aeron fragments with no length check
(`networking/src/server/cmd_handler.rs:13`), and the SBE accessors are raw slice indexes:

```rust
slice[index..index + N].try_into().expect("slice with incorrect length")   // order/src/lib.rs:80
```

Measured in a release build: **every buffer length from 0 to 63 panicked**; only 64 decoded. The
panic fires inside a callback invoked from Aeron's C poll loop through `rusteron`'s `extern "C"`
trampoline, and unwinding out of an `extern "C"` frame aborts the process.

The SBE header was also never validated in release — the only check was a `debug_assert_eq!` on the
template id (`order_command_message_codec.rs:267`), compiled out in release, and the schema id was
never read at all. Corrupting the first four bytes still decoded cleanly.

Combined with the unauthenticated gateway handshake (#114), this was a remote crash.

## The fix

A single guard at the decode boundary, before any field is read:

- reject buffers shorter than `ORDERCOMMANDSIZE`;
- reject headers whose `block_length`, `template_id` or `schema_id` are not the expected values.

Both return a descriptive `Err` through the function's existing error type, so every caller already
handles it — no call site changes and no `catch_unwind`.

This is three integer comparisons on the ingress hot path. The per-field accessors are deliberately
left as raw indexes: making them individually fallible would add a branch per field to every decode
for no additional safety once the boundary is checked.

## Cross-reference

`vex-gateway` has the identical construction on its side of the link, decoding engine responses with
no length check (trade-vex/vex-gateway#59). Because the codec is shared via the `common` crate, this
fix covers both directions.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p common -p sbe_order`
passes, with new tests asserting every length 0..63 returns `Err` without panicking, a valid message
round-trips, and a corrupted header is rejected.

Closes #115

### DIFF
--- a/common/src/cmd.rs
+++ b/common/src/cmd.rs
@@ -2,10 +2,10 @@ use crate::{L2MarketData, OrderCommandType, Side, TimeInForce, UserBalance};
 use borsh::{BorshDeserialize, BorshSerialize};
 use sbe_order::message_header_codec::{self, MessageHeaderDecoder};
 use sbe_order::order_command_message_codec::{
-    OrderCommandMessageDecoder, OrderCommandMessageEncoder,
+    OrderCommandMessageDecoder, OrderCommandMessageEncoder, SBE_BLOCK_LENGTH, SBE_TEMPLATE_ID,
 };
 use sbe_order::status::Status as SbeStatus;
-use sbe_order::{ReadBuf, SbeResult, WriteBuf};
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #148, opened 2026-08-14
fix: reject a cancel that does not come from the order's owner

rahulghangas:
## The problem

`cancel_order` cancelled any order id it was given, with no check that the requester owned it.
Worse, `PriceLevel::remove_order` **overwrote** the requester with the resting order's owner:

```rust
cmd.set_user_id(removed_order.user_id);   // orderbook/src/lib.rs:80
```

so Risk R2 then refunded the *victim* — which made an unauthorised cancel look legitimate all the
way downstream. Order ids are structurally predictable (`timestamp << 16 | sequence << 4 | gateway`),
so guessing a live id is not hard.

## The fix

Compare before mutating. The requesting user is already on the command — the gateway sets
`OrderCommand.user_id` to the authenticated user when it builds a `CancelOrder` — so the book only
needed to check it instead of discarding it:

- `PriceLevel::remove_order` now compares `cmd.user_id` against the resting order's owner and
  rejects on mismatch, leaving the level, the volume and the order untouched.
- `OrderBook::cancel_order` no longer removes the `orders` index entry up front; it removes it only
  once the cancel has actually succeeded. Previously a failed lookup dropped the index entry and
  orphaned the order permanently.

`remove_order` is private, so returning `bool` is not an API change. On the success path
`set_user_id` is now a no-op (requester == owner) and is left in place so the R2 sharding and refund
behaviour are byte-for-byte unchanged.

The check is one integer comparison with no allocation, and the removal is written without `unwrap`
so it adds no panic site to the matching thread.

## Correction: the requester was not on every cancel

The paragraph above — "the requesting user is already on the command" — was true of the gateway and
of nothing else, and this PR now carries a second commit to fix that.

The gateway's handler does populate the field from the authenticated identity
(`crates/libs/lib-api/src/handlers/order.rs`). But `OrderCommand::cancel_order`, the constructor
every in-repo caller uses, hardcoded it:

```rust
pub fn cancel_order(order_id: u64, side: Side, market_id: u32) -> Self {
    Self { /* ... */ user_id: 0, /* ... */ }
}
```

and the field's own doc comment described `user_id` on a `CancelOrder` as deliberately holding a
*gateway* id — "user_id is not essential for the CancelOrder hence it is the best choice to place
gateway id to send response correctly". The ownership guard turns that field from an output into a
**required input**, so every cancel constructed inside this repo arrived with `user_id = 0` and was
rejected as a third-party cancel:

```
event="command_processed" … user_id=0 command=CancelOrder status=Rejected stage="events"
```

`cargo check` and clippy were silent about it — `user_id: 0` is perfectly well-typed — so the
follow-up makes `user_id` a **positional argument** on `cancel_order` rather than something a caller
can forget. That converts every stale call site from a runtime rejection into a compile error, which
is the only reliable way to find them all. There were more than expected:

- `server/src/lib.rs` — 2 sites, both in tests that were already failing (see Verification).
- `xtask/src/scenarios/cancellation.rs` — 9 sites; `xtask/src/scenarios/comprehensive.rs` — 3 sites.
  `xtask`'s typestate `CancelBuilder` had no `user_id` field at all, so `OrderBuilder::cancel()` now
  takes the caller and threads it through each state transition to `build()`.

This is not a wire or protocol change: `user_id` was already on `OrderCommand` and the gateway was
already filling it in. What changed is the in-repo constructor, its callers, and the doc comment,
which now states what the field means — owner on a place, caller on a cancel — and points response
routing at `route_gateway_id`, which is a separate field and always was.

## Cross-reference

The gateway does not check ownership either (trade-vex/vex-gateway#53). The engine is the right
place for the authoritative check: the order book is the only component with synchronous truth about
who owns a resting order. The gateway's own view is populated asynchronously from Kafka, so a
gateway-side pre-check would reject legitimate cancels of just-placed orders while adding nothing
this fix does not already guarantee.

## Verification

The verification originally claimed here stopped at `cargo test -p vex-orderbook`, and that was not
enough. On the guard commit alone, `cargo test -p vex-server --lib` failed in both tests that cancel
an order:

```
thread 'tests::test_multi_market_stress_scenario' panicked at server/src/lib.rs:1778:9:
assertion `left == right` failed
  left: Rejected
 right: Cancelled

failures:
    tests::test_complex_scenario_with_cancellations_and_mixed_tif
    tests::test_multi_market_stress_scenario

test result: FAILED. 5 passed; 2 failed
```

Both of those tests cancel an order they placed themselves, so both are legitimate owner cancels;
they were rejected purely because the constructor supplied `user_id = 0`. Passing the caller in
fixes them without touching either assertion's expectation.

On the head of this branch:

- `cargo test -p vex-orderbook` — 75 passed, 0 failed, including new cases asserting a third-party
  cancel is rejected with the order still on the book and still cancellable by its true owner.
- `cargo test -p vex-server --lib` — 7 passed, 0 failed.
- `cargo clippy --workspace --all-targets` — 0 warnings.

Closes #121

### DIFF
--- a/common/src/cmd.rs
+++ b/common/src/cmd.rs
@@ -46,14 +46,9 @@ pub struct OrderCommand {
     pub order_id: u64,
 
     /// The unique identifier of the user initiating the command.
-    /// For 'PlaceOrder' : This is the actual user id
-    /// For 'CancelOrder': this is set to the gateway id inside the OrderCommand Field
-    /// This is required because, in the existing scenarior the gateway id(required to send
-    /// back the response to gateway with gateway_id in the events handler) not present in
-    /// the OrderCommand Field, so it is encoded in the order_id by the snowflake algorithm
-    /// Everything works correctly untill the order_id is itself wrong that is sent the gateway
-    /// and user_id is not essential for the CancelOrder hence it is the best choice to place
-    /// gateway id to send response correctly.
+    /// For `PlaceOrder`, this is the owner of the new order.
+    /// For `CancelOrder`, this is the caller whose ownership of the existing order is checked.
+    /// Gateway response routing uses `route_gateway_id`, not this field.
     pub user_id: u64,
 
     /// The identifier for the market this command targets.
@@ -153,12 +148,12 @@ impl OrderCommand {
         }
     }
 
-    pub fn cancel_order(order_id: u64, side: Side, market_id: u32) -> Self {
+    pub fn cancel_order(order_id: u64, user_id: u64, side: Side, market_id: u32) -> Self {
         Self {
             command: OrderCommandType::CancelOrder,
             order_id,
             market_id,
-            user_id: 0,
+            user_id,
             price: 0,
             size: 0,
             side,
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -70,23 +70,30 @@ impl PriceLevel {
 
     #[inline]
     fn remove_order(&mut self, order_id: u64, cmd: &mut OrderCommand) -> bool {
+        let requesting_user_id = cmd.user_id;
         if let Ok(pos) = self
             .orders
             .binary_search_by_key(&order_id, |order| order.order_id)
-            && let Some(removed_order) = self.orders.remove(pos)
         {
-            self.total_volume -= removed_order.size;
-            cmd.set_price(removed_order.price);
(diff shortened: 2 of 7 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #151, opened 2026-08-14
fix: validate the native scale factors that actually drive collateral

rahulghangas:
## The problem

`SymbolSpecificationConfig::validate` guarded the wrong field.

It checks `base_scale_k != 0` — a field that is **never used anywhere in the workspace** — while
`base_native_scale` and `quote_native_scale`, which form the live collateral divisor, were not
validated at all:

```rust
let denominator = (self.base_native_scale as u128) * (self.quote_scale_k as u128);
if denominator == 0 { return 0; }        // common/src/market_specification.rs:41
```

A zero `base_native_scale` therefore made `calculate_quote_cost` return 0 for that market — pricing
**every trade in it at zero** — and configuration validation passed happily.

Separately, `production_defaults()` ships an empty symbol map and an empty map passes `validate()`
trivially, so a production instance could start with no markets at all.

## The fix

- Reject `base_native_scale == 0` and `quote_native_scale == 0`, in the same style as the existing
  checks.
- Reject an empty symbol set when the environment is Production.

The emptiness check lives in `VexConfig::validate`, which already owns `self.environment`, rather
than threading an environment parameter through `SymbolSpecificationConfig::validate` and breaking
its signature — same result, smaller change.

The existing `base_scale_k` check is left in place; its deadness is tracked separately in #145.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p vex-config` passes
37 tests, including new cases for each zero native scale, an empty symbol set rejected in Production
and accepted in Development, and a valid config still accepted. One production fixture in
`tests/integration_tests.rs` needed real symbols to satisfy the new rule.

Closes #133

## Summary by CodeRabbit

* **New Features**
  * Added a development configuration with networking, logging, market, and Kafka settings.
  * Added support for selecting environments through `VEX_ENV`, `ENVIRONMENT`, `ENV`, or `NODE_ENV`.
  * Added clearer configuration load errors with searched paths and underlying causes.

* **Bug Fixes**
  * Prevented production configurations from using empty market symbols or zero native scales.
  * Ensured required runtime libraries are available in Docker deployments.

* **Documentation**
  * Updated runbooks and Docker guidance to require explicit environment selection.

### DIFF
--- a/vex-config/src/lib.rs
+++ b/vex-config/src/lib.rs
@@ -122,6 +122,11 @@ impl VexConfig {
         self.gateway_networking.validate()?;
         self.logging.validate()?;
         self.symbols.validate()?;
+        if self.is_production() && self.symbols.is_empty() {
+            return Err(ConfigError::ValidationError(
+                "Symbol configuration cannot be empty in production".to_string(),
+            ));
+        }
         #[cfg(feature = "balance-preload")]
         if self.balance_preload.enabled {
             self.balance_preload.validate()?;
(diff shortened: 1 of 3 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #153, opened 2026-08-14
fix: make trade settlement atomic and surface its failures

rahulghangas:
## The problem

`settle_trade` performed up to three fallible balance mutations in sequence, each with `?`, under a
single lock and with **no rollback**:

```rust
store.unlock_funds(user_id, quote_asset(market_id), refund_amount)?;      // :231 price improvement
let balance_sub = store.subtract_locked_funds(user_id, sub_asset, sub_amt)?;  // :267
let balance_add = store.add_funds(user_id, add_asset, add_amt)?;              // :269
```

If step 2 or 3 failed, everything already applied stayed applied. The matching engine has already
executed the trade and mutated the book by this point, so a failure left a **half-applied trade** and
total supply changed.

The caller then swallowed it — `error!` and return (`:180-197`).

Measured: with a ledger divergence in place, the taker was credited 100 base and debited 10,000
quote while the maker received nothing.

## The fix

Compute-then-apply. Since everything already happens under one lock and `UserBalance` is `Copy`,
every delta is computed and validated against copies first, and the store is written only once all
of them are known to succeed. No transaction abstraction, no rollback log, no allocation.

The same-asset case (`asset_to_add == asset_to_subtract`) is handled explicitly so the two writes
cannot clobber each other.

On failure the store is not touched at all, and the `error!` now carries the full settlement context
— user, market, side, maker, price, size, taker limit price — instead of just the error.

### Why the command is not marked `Rejected`

I checked what the R2 stage and the events handler do with a `Rejected` status before choosing.
`server/src/utils.rs`'s R2 handler returns early on `Rejected`, so marking would cause the *other*
shards to skip settlement for the remaining events in the chain, and `processors/src/events.rs`
suppresses trade publication for `Rejected` — so marking would turn a partial failure into a larger,
quieter one. The status is therefore left alone and the failure is surfaced through the log.

Making a settlement failure genuinely fatal (halting the market) is the right end state but is a
larger behavioural change than this fix, and is worth deciding separately.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p processors -p common`
passes, including a new regression test asserting that when the subtract step fails, **both**
participants' balances are byte-for-byte what they were before the call.

Closes #113

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -184,6 +184,11 @@ impl RiskEngine {
                 shard_id = self.shard_id,
                 user_id,
                 market_id,
+                user_side = ?user_side,
+                maker_user_id = event.maker_user_id,
+                price = event.price,
+                size = event.size,
+                taker_price = ?taker_cmd,
                 error = ?err
             );
         } else {
@@ -213,12 +218,9 @@ impl RiskEngine {
             spec.taker_fee
         };
 
-        // Acquire a lock on the store for all balance operations
-        let mut store = self.balances.lock();
-
         // --- Price Improvement Refund Logic (for Taker only) ---
         // If the taker gets a better price than their limit, refund the difference.
-        if !is_maker
+        let refund_amount = if !is_maker
             && let Some(limit_price) = taker_price
             && user_side == Side::Bid
         // Price improvement only applies to BID orders where QUOTE currency was locked.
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #154, opened 2026-08-14
fix: make fee arithmetic overflow-safe and bound the fee configuration

rahulghangas:
## The problem

Two defects in the same pair of lines in `settle_trade`, on both the Bid and Ask arms:

```rust
let fee_in_base = (event.size * fee) / 10000;      // bare u64 multiply
let base_received_net = event.size - fee_in_base;  // bare u64 subtract
```

**Overflow.** For an 18-decimal base asset one unit is `size = 1e18`, and `1e18 * 20 = 2e19` exceeds
`u64::MAX` (1.845e19). The release profile has `overflow-checks` off, so it wrapped:

| size | fee | correct fee | actual fee (release) |
|---|---|---|---|
| 1e18 | 20bp | 2,000,000,000,000,000 | 155,325,592,629,044 |

**92.2% of the fee silently lost.** In debug it panicked instead, which wedges the pipeline.

**Underflow.** `size - fee_in_base` had no upper bound on the fee. `validate()` only checked
`taker_fee >= maker_fee`, so `taker_fee = 20000` — a plausible "2%" typo — made `fee > size`:

```
size=1000 taker_fee=20000bp -> fee_in_base=2000 > size -> net = 18446744073709550616
```

## The fix

- Compute both fee amounts through `u128` and narrow back saturating, matching the idiom already
  used by `calculate_quote_cost`. The multiply can no longer wrap.
- Make the net-amount subtraction saturating, so it cannot underflow even if a fee were somehow
  out of range.
- Reject `taker_fee >= 10000` and `maker_fee >= 10000` in `vex-config`'s `validate()`, in the style
  of the existing checks. This is a bound on an existing field, not a new setting.

Fee semantics are unchanged — still basis points, still floor division.

## Note

This makes the fee *correct*; it does not change where the fee goes. Fees are still deducted from
the credited amount and credited to no account, so total supply still shrinks on every trade. That
is tracked separately in #118.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p processors -p common
-p vex-config` passes 69 tests, including a new settlement test at 18-decimal scale asserting the
mathematically correct fee rather than the wrapped one, and a config test rejecting a fee at the
10000bp boundary.

Closes #119

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -242,8 +242,9 @@ impl RiskEngine {
             Side::Bid => {
                 let quote_spent = spec.calculate_quote_cost(event.price, event.size);
                 // Fee is on the base asset received. Assuming fee is in basis points (e.g., 20bp = 0.2%)
-                let fee_in_base = (event.size * fee) / 10000;
-                let base_received_net = event.size - fee_in_base;
+                let fee_in_base =
+                    u64::try_from((event.size as u128 * fee as u128) / 10000).unwrap_or(u64::MAX);
+                let base_received_net = event.size.saturating_sub(fee_in_base);
                 (
                     quote_asset(market_id),
                     quote_spent,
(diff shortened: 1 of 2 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #155, opened 2026-08-14
fix: reject orders instead of dropping them when the ring buffer is full

rahulghangas:
## The problem

When the disruptor ring was full, the order was thrown away:

```rust
if let Err(e) = self.producer.try_publish(|cmd| { *cmd = order_command.clone(); }) {
    error!(target: "gateway_fragment", ..., "failed to publish order command to ring buffer");
}
// cmd_handler.rs:24-33 — and that is all that happens
```

The client was never told: no rejection, no NACK. It was not archived either, because journaling is
a pipeline stage a dropped command never reaches. On the gateway side this surfaced as a 5-second
timeout returning HTTP 408 with `order_id 0`, so a dropped order and a dropped response are
indistinguishable to a client — it cannot know whether to retry.

It was also easy to hit: `Duologue::poll` used a fragment limit of **2048** against a **1024**-slot
ring, so a single poll could hand the ring twice its capacity.

## The fix

- On `try_publish` failure, mark the command `Status::Rejected` and send it back to the originating
  gateway via the existing `Publications::publish_response`. `route_gateway_id` is already set
  earlier in the handler, so it routes correctly. The client now gets a definitive answer instead of
  a timeout.
- Lower the poll fragment limit to the ring capacity so one poll cannot overrun it 2×.

The ingress thread still never blocks and never retries — it is the single Aeron poll thread for
every gateway, so stalling it would be worse than the problem being fixed.

## Note on the constant

`RING_BUFFER_CAPACITY` duplicates `vex_server::engine::BUFFER_SIZE` rather than importing it,
because `server` depends on `networking` and not the reverse. It carries a doc comment saying so.
Hoisting the value into `common` would be the cleaner end state but is a wider change than this fix
warrants.

## No unit test

Observing the response requires a concrete `AeronPublication`, so a test without a live Aeron media
driver would assert nothing meaningful. Stated plainly rather than adding a vacuous one.

## Cross-reference

The gateway-side symptom — a 408 with no idempotency key and no reconciliation — is
trade-vex/vex-gateway#54. This change removes the ambiguity for the ring-full case specifically.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p vex-networking -p
common` passes.

Closes #123

### DIFF
--- a/networking/src/server/cmd_handler.rs
+++ b/networking/src/server/cmd_handler.rs
@@ -1,11 +1,14 @@
+use crate::server::gateway_publications::Publications;
 use common::{OrderCommand, Status, decode_order_command};
 use disruptor::{MultiProducer, Producer, SingleConsumerBarrier};
 use rusteron_archive::{AeronFragmentHandlerCallback, AeronHeader};
+use std::sync::Arc;
 use tracing::{debug, error, info};
 
 pub struct FragmentHandler {
     pub gateway_id: u8,
     pub producer: MultiProducer<OrderCommand, SingleConsumerBarrier>,
+    pub publications: Arc<Publications>,
 }
 
 impl AeronFragmentHandlerCallback for FragmentHandler {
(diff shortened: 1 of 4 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #156, opened 2026-08-14
fix: abort loudly when a pipeline handler panics instead of wedging silently

rahulghangas:
## The problem

Each disruptor stage runs on a bare thread. If a handler panics, that thread dies, its consumer
cursor stops advancing, every downstream barrier blocks forever and the ring fills. Nothing
supervises the threads — `Consumer::join`'s `.expect("Consumer should not panic.")` only runs on
drop, which never happens in practice.

Demonstrated with a faithful replica of the topology, one handler panicking at event 5:

```
published 64 of 400 before the pipeline stalled
stage A healthy handler saw 64 events
stage B (downstream of the barrier) saw 0 events
ring buffer permanently FULL
```

To an operator this is invisible: the Docker healthcheck is process-liveness only, there are no
metrics, and the only symptom is an endless stream of "failed to publish order command to ring
buffer".

## The fix

One helper, `abort_on_handler_panic(stage, handler)`, wrapping all 14 production pipeline handlers.
A panic is logged at `error!` with the stage identity and the sequence number, then the process
aborts.

### Why abort rather than a graceful shutdown flag

Once a stage thread has died the pipeline **cannot** drain — downstream barriers never advance — so
signalling the existing shutdown flag would hang rather than shut down. The flag is also only
constructed after the pipeline is built, so a panic during early operation could not reach it. A
fast, loud death that an orchestrator can see (SIGABRT, exit 134) is strictly better than the
current silent, permanent stall.

The trade-off is that an abort skips the Kafka producer flush in `KafkaEventsHandler::drop`. That is
acceptable: the alternative is an exchange that is wedged and looks healthy.

`catch_unwind` costs nothing on the success path — it only emits landing pads — so the hot path is
unaffected.

## Verification

`cargo check` and `cargo clippy --workspace --all-targets` clean; `cargo test -p vex-server -p
processors` passes, including a subprocess test asserting a panicking handler logs its stage and
terminates the child by signal rather than stalling.

Closes #129

### DIFF
--- a/server/src/engine.rs
+++ b/server/src/engine.rs
@@ -15,11 +15,12 @@ use processors::{
 };
 use std::{
     cell::UnsafeCell,
+    panic::{AssertUnwindSafe, catch_unwind},
     sync::Arc,
     sync::atomic::AtomicBool,
     thread::{self, JoinHandle},
 };
-use tracing::info;
+use tracing::{error, info};
 use vex_config::{CoreNetworkingConfig, GatewayAuthenticationKey};
 use vex_networking::server::Publications;
 use vex_networking::server::VexCoreServer;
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

Issue #80, opened 2025-10-22
fix: cancel should use linear scan


rahulghangas:
I think cancel should use a linear scan instead of binary search on VecDeque. Need to confirm behaviour

rahulghangas:
Related: #122. The linear-scan change tracked here would also fix a correctness bug: because the `orders` index stores only a price and `cancel_order` probes `bids` before `asks`, cancelling an order at a price where the opposite side also has a level (reachable via self-trade prevention) removes the index entry, fails the lookup, and orphans the order permanently with its funds locked.

PR #53, opened 2025-08-28
Feat/symbol spec

aditya-git16:
## Summary by CodeRabbit

- New Features
  - Aeron-based gateway client/server, with SBE encode/decode for order commands.
  - Core engine with journaling, risk, and matching shards; symbol management.
  - Dockerized setup and docker-compose for media driver, codegen, and app.
  - Config system with env-aware loading, validation, and examples.
  - Order book and models updated to Side-based, u64 types.
  - xtask CLI for e2e tests and benchmarks.

- Documentation
  - Docker setup guide; multiple configuration examples.

- Tests
  - Client–server integration, server e2e, and updated order book tests.

- Chores
  - Workspace expansion; ignore rules and aliases.

PR #157, opened 2026-08-14
fix(orderbook): make the cancel index side-aware so a shared price cannot orphan an order

rahulghangas:
## The problem

`OrderBook::cancel_order` resolved a resting order from `orders: HashMap<u64, u64>`, which maps
`order_id -> price` only. It then removed the index entry **first** and probed `bids`, falling back
to `asks`:

```rust
if let Some(price) = self.orders.remove(&cmd.order_id) {
    if let Some(level) = self.bids.get_level_mut(price) { ... }
    else if let Some(level) = self.asks.get_level_mut(price) { ... }
```

`cmd.side` was ignored — and it is unreliable for a cancel anyway, since the gateway does not know
the resting side.

When both sides hold a level at the same price, an **ask** resolves to the **bid** level, the
binary search inside `remove_order` misses, and the cancel is rejected. Because the index entry was
already removed, the retry is rejected too: the order rests forever and the user's funds stay
locked with no way to release them.

That state is reachable in normal operation. Self-trade prevention makes the matcher `continue`
past a user's own resting order, so the aggressor rests instead of crossing — leaving a level on
each side at the same price.

## The fix

- `orders` becomes `HashMap<u64, (u64, Side)>`. The side is recorded at `add_to_book`, the single
  insertion point, so a cancel goes directly to the correct side instead of probing.
- The index entry is now removed **only after** the level removal succeeds, so a failed cancel is
  retryable rather than self-destructive.
- `PriceLevel::remove_order` returns `bool` so `cancel_order` can tell the two apart.

`Side` is a one-byte enum, so the index value grows from 8 to 16 bytes with alignment — one entry
per resting order. Cancel goes from up to two level probes to exactly one.

## Tests

`orderbook/src/unit_tests.rs` gains coverage for a bid and an ask resting at the same price (each
cancellable by its owner, book empty afterwards) and for ordinary single-sided cancels on both
sides. The `verify_state` invariant checker is now side-aware, so every existing test asserts the
index side matches the level the order actually sits in.

`cargo test -p vex-orderbook`: 75 passed, 0 failed. clippy: 0 warnings.

## Related

Closes #122

- vex-core #80 — earlier report of the same cancel path (suggested a linear scan; this fix keeps
  the O(1) index and makes it side-aware instead).
- vex-core #148 — adds the ownership check inside `PriceLevel::remove_order`; touches the same
  function signature, so whichever merges second will need a trivial rebase.
- vex-gateway #53 — closed as addressed engine-side; the gateway cannot validate a cancel because
  resting orders only reach Postgres via async Kafka.

### DIFF
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -15,8 +15,9 @@
 //!     -   **Order Queue (`VecDeque`):** Within each `PriceLevel`, a `VecDeque` stores the orders.
 //!         This acts as a FIFO (First-In, First-Out) queue, ensuring time priority for orders
 //!         at the same price.
-//!     -   **Direct Order Access (`HashMap`):** A `HashMap<u64, u64>` provides O(1) average-case
-//!         lookup time for order-price pairs by their ID. This is crucial for fast cancellation.
+//!     -   **Direct Order Access (`HashMap`):** A `HashMap<u64, (u64, Side)>` provides O(1)
+//!         average-case lookup time for order price and side by ID. This is crucial for fast
+//!         cancellation.
 //!
 //! 2.  **Performance Characteristics:** [Depends on BookSide Implementation]
 //!     [For BTreeMap-based BookSide]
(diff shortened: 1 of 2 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #52, opened 2025-08-28
Test/risk engine

aditya-git16:
## Summary by CodeRabbit

- New Features
  - End-to-end gateway ↔ core networking with Aeron, handshake, and order command streaming.
  - Configurable environments (Development/Test/Production) with validation and merging.
  - Dockerized setup (media driver, SBE generator, core) with compose scripts and health checks.
  - Core engine with sharded matching/risk processing and journaling.
  - CLI tooling (xtask) for e2e tests and benchmarks.

- Refactor
  - Unified “Side” semantics and unsigned IDs across commands, models, and order books; streamlined command set.

- Documentation
  - Added Docker setup guide and configuration examples.

- Tests
  - Integration and e2e tests for client/server, networking, and configuration.

- Chores
  - Workspace expansion, ignores, and build aliases.

PR #162, opened 2026-08-15
fix(client): add a non-blocking order send so callers need not sleep 3.1s on back-pressure

### LINKED ISSUE #161: fix: the gateway client's only send path blocks the calling thread for 3.1s under back-pressure
## Summary

`Publisher::send_order_command` (`networking/src/client/mod.rs:~286`) is the only way to publish an
`OrderCommand`, and it retries an Aeron `offer` five times with `std::thread::sleep`:

```rust
for attempt in 0..MESSAGE_RETRY_COUNT {
    let result = self.publication.offer::<AeronReservedValueSupplierLogger>(&message_buffer, None);
    if result >= 0 { return Ok(()); }
    let delay = MESSAGE_RETRY_DELAY * (2_u32.pow(attempt as u32));
    std::thread::sleep(delay);          // 100/200/400/800/1600 ms
}
```

That is **3.1 seconds of blocking sleep**, and there is no alternative entry point. A caller that
cannot afford to block — an async runtime, or anything holding a lock — has no way to ask "can I
publish right now?" and act on the answer.

## Impact

`vex-gateway` calls this from inside an actix async handler while holding a process-global
`tokio::Mutex` (trade-vex/vex-gateway#52). One back-pressured order therefore:

1. blocks an entire actix **worker thread** for 3.1 s, starving every other request assigned to that
   worker — not just orders; and
2. serialises every other order in the process behind the mutex.

Aeron back-pressure is exactly the condition that arises under load, so this degrades hardest
precisely when it matters. The gateway wants to answer HTTP 503 immediately, and cannot.

## Suggested fix

Add a non-blocking `try_send_order_command` that offers once and returns immediately, with an error
type that distinguishes a transient Aeron condition (`NOT_CONNECTED`, `BACK_PRESSURED`,
`ADMIN_ACTION`) from a genuine failure. Keep `send_order_command` and its behaviour for existing
callers, implemented on top of the same primitive.

The server side already makes this distinction in
`networking/src/server/gateway_publications.rs`; the client should not disagree with it about what
counts as transient.

## Related

- trade-vex/vex-gateway#52 — the caller-side half.
- #116, #126 — the same offer-result classification on the archive and response paths.

rahulghangas:
## The problem

`Publisher::send_order_command` was the **only** way to publish an `OrderCommand`, and it retries an
Aeron `offer` five times with `std::thread::sleep` at 100/200/400/800/1600 ms — 3.1 seconds of
blocking sleep. A caller that cannot afford to block had no alternative and no way to ask whether a
publish would succeed right now.

`vex-gateway` calls this from an actix async handler while holding a process-global `tokio::Mutex`,
so a single back-pressured order blocks an entire worker thread for 3.1 s *and* serialises every
other order behind the mutex. Back-pressure is exactly what happens under load.

## The fix

Additive. `send_order_command` keeps its exact signature, retry count and backoff timing — the
gateway pins this crate by git rev and other callers depend on it.

- **`try_send_order_command`** encodes once, offers once, returns immediately. No loop, no sleep.
- **`GatewayError::Unavailable`** is the transient case, covering Aeron's `NOT_CONNECTED`,
  `BACK_PRESSURED` and `ADMIN_ACTION`. All three mean "the engine cannot take this right now, try
  again", which is one decision for the caller — answer 503 — so it is one variant, documented with
  the codes it covers. A closed publication or an unrecognised code stays a `SendError`.
- Both public functions share one private `offer_order_command` helper, so the retry path and the
  single-shot path cannot drift apart. `send_order_command` still encodes **once** before the loop
  and offers the same bytes each attempt.

`ADMIN_ACTION` being transient matters: it is a routine term rotation. Classifying it as a hard
failure would have the gateway report HTTP 500 for an event that resolves by itself in microseconds.
This matches how `networking/src/server/gateway_publications.rs` already classifies offer results on
the archive and response paths — the client and the server now agree on what "transient" means, using
the same named `rusteron` constants rather than raw integers.

`cargo test -p vex-networking`: 7 passed, 0 failed. clippy: 0 warnings.

## Scope

This is the engine half only. Nothing in `vex-core` calls the new function yet — it exists so the
gateway can stop blocking. The caller-side change is vex-gateway#52 and needs the vex-core pin in
`vex-gateway/Cargo.toml` (currently `rev = "efeaeb4"`) moved to a commit containing this, so this
must merge first.

## Related

Closes #161

- vex-gateway #52 — the caller-side half: publish without blocking a worker, and return 503 instead
  of sleeping.
- vex-core #116 / PR #159 and #126 / PR #160 — the same offer-result classification applied to the
  archive and client-response paths.

## Summary by CodeRabbit

* **New Features**
  * Added a non-blocking option for sending order commands.
  * Improved handling of temporarily unavailable publication channels with automatic retries.

* **Bug Fixes**
  * Protocol errors now return immediately with clearer error reporting.
  * Added handling for closed publications and unexpected publication results.

### DIFF
--- a/networking/src/client/mod.rs
+++ b/networking/src/client/mod.rs
@@ -5,6 +5,10 @@ use crate::utils::{
 use common::ORDERCOMMANDSIZE;
 use common::{OrderCommand, encode_order_command, order_debug};
 use rand;
+use rusteron_archive::bindings::{
+    AERON_PUBLICATION_ADMIN_ACTION, AERON_PUBLICATION_BACK_PRESSURED,
+    AERON_PUBLICATION_NOT_CONNECTED,
+};
 use rusteron_archive::{
     Aeron, AeronCError, AeronContext, AeronFragmentHandlerCallback, AeronHeader, AeronPublication,
     AeronReservedValueSupplierLogger, AeronSubscription, Handler,
@@ -238,6 +242,10 @@ pub enum GatewayError {
     CoreError(String),
     #[error("Failed to send message: {0}")]
     SendError(String),
+    /// Publication is temporarily unavailable due to Aeron `NOT_CONNECTED`, `BACK_PRESSURED`, or
+    /// `ADMIN_ACTION`.
+    #[error("Aeron publication is temporarily unavailable")]
+    Unavailable,
     #[error("Protocol error: {0}")]
     ProtocolError(String),
     #[error("Configuration error: {0}")]
@@ -285,7 +293,6 @@ impl Publisher {
 
     /// Sends an OrderCommand to the core
     pub fn send_order_command(&self, order_command: &OrderCommand) -> Result<(), GatewayError> {
-        // Send the binary message directly
         order_debug!(
             "gateway_send_attempt",
             order_command,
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #163, opened 2026-08-15
fix(risk): round fees up so slicing an order is no longer free

rahulghangas:
> Stacked on #154 (`fix/119-fee-arithmetic`), whose overflow-safe form this builds on. Merge that
> first.

## The problem

Fees are computed as `(size * fee) / 10000` in basis points. Integer division **floors**, so any
fill below `10000/fee` units pays exactly zero. At a 20bp taker fee that is any fill under 500 units.

Nothing stops a client from slicing:

```
20 slices x 499 units, taker fee 20bp   -> total fee paid = 0
one 9,980-unit order                    -> floor(9980*20/10000) = 19
```

Trading fee-free is a matter of choosing your fill size.

## The fix

Round the fee **up** — `div_ceil(10000)` on both the base-side and quote-side computations.

Boundaries, all covered by new tests:

| Case | Before | After |
|---|---|---|
| rate 0 (maker fee is legitimately 0 in the shipped config) | 0 | 0 |
| exact multiple, e.g. 20,000 units @ 20bp | 40 | 40 |
| 20 × 499 units @ 20bp | 0 total | 1 each, 20 total |
| 1 unit @ 20bp | 0 | 1 |

A zero-fee market stays free — `div_ceil` of zero is zero — so this cannot accidentally start
charging where no fee was configured.

`div_ceil` rather than a hand-rolled `(a + b - 1) / b`, which can overflow. The multiply stays
widened to `u128` and the subtraction stays saturating, exactly as #154 left them.

## Correction: existing test expectations *did* change

This description originally claimed **"No existing test expectation changed."** That was wrong.

The audit behind that claim only covered `processors/src/risk_engine.rs`, where every hard-coded fee
happens to be an exact multiple of 10000 — taker 20,000 @ 20bp = 40, maker 1,000,000,000 @ 10bp =
1,000,000, the 18-decimal fill = 2,000,000,000,000,000, the 1,000-unit fill @ 20bp = 2 — so ceiling
and floor agree there. It never looked at the callers in other crates, and one of them was red:

```
thread 'tests::test_complex_scenario_with_cancellations_and_mixed_tif'
panicked at server/src/lib.rs:1266:9:
assertion `left == right` failed
  left: 27
 right: 30
```

`5a6b931` recomputes those expectations **per fill**, which is the semantics of the fix rather than a
concession to it — the engine applies the rate to each fill, not to the order total, and that is
precisely why slicing no longer helps.

**`vex-server`** — `test_complex_scenario_with_cancellations_and_mixed_tif`. Taker D's 30-unit bid is
filled by three makers: 10 from A, 15 from B, 5 from C. The test encoded the aggregate floor form,
`gross_base_recv - (gross_base_recv * 20 / 10000)` = 30 − 0 = 30. Each of the three fills now rounds
up to 1 BASE, so D nets 27. The maker side moved the same way — A's 10bp on 10,100 quote goes 10 →
11, B's on 15,300 goes 15 → 16, C's on 5,100 goes 5 → 6 — and those asserts now derive the fee with
`div_ceil` instead of restating a floor result.

That test also carried a comment claiming D took 25 units as taker and 5 "as maker (from E)". That
was already false: Taker E's IOC ask is cancelled against an empty book after C cancels, so all 30
units reach D as taker. Under floor rounding every fee was 0, so the wrong narrative still produced
the right number and nothing caught it. Corrected alongside.

**`xtask`** — two IOC scenarios did not merely miss the defect, they asserted it:

- `test_market_buy_section`, 4 units @ 10bp — expected fee 0, now 1.
- `test_market_sell_section`, 10 units @ 10bp — expected fee 0, now 1. Its comment read
  `fee_btc = (10 * 10) / 10000 = 0 (rounds down)`.

So the original description reached the right conclusion from the wrong evidence: the suite was not
blind to this defect everywhere — in places it was pinning the defect in place.

## Verification

`cargo test -p processors`: 19 passed, 0 failed. clippy: 0 warnings.

CI **Build and Test** failed on `00ad7df` and passes on `5a6b931` — `processors` 19 passed / 0
failed, `vex-server` lib 7 passed / 0 failed.

Clippy, End-to-end tests and Security audit are red on this branch and equally red on `develop` at
the merge base (`2852560`): Clippy on a `format!` lint in `vex-networking`, E2E on a `vex-networking`
build failure against `rusteron-archive` (`AeronArchiveContext::new_with_no_credentials_supplier` no
longer exists). Nothing here touches either. The consequence for review is that the two xtask
scenario corrections are right in source but were not exercised by CI on this branch.

## Scope

This makes the fee non-zero. It does **not** change where the fee goes — it is still debited from
the user and credited to no account, which is #118.

## Related

Closes #120

- vex-core #119 / PR #154 — overflow-safe fee arithmetic and the `taker_fee < 10000` bound this
  relies on; stacked on it.
- vex-core #118 — fees are debited and credited nowhere. Rounding up increases the amount that
  currently vanishes.
- vex-core #111 / PR #146 — the related rounding-to-zero defect on the collateral side.

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -242,8 +242,8 @@ impl RiskEngine {
             Side::Bid => {
                 let quote_spent = spec.calculate_quote_cost(event.price, event.size);
                 // Fee is on the base asset received. Assuming fee is in basis points (e.g., 20bp = 0.2%)
-                let fee_in_base =
-                    u64::try_from((event.size as u128 * fee as u128) / 10000).unwrap_or(u64::MAX);
+                let fee_in_base = u64::try_from((event.size as u128 * fee as u128).div_ceil(10000))
+                    .unwrap_or(u64::MAX);
                 let base_received_net = event.size.saturating_sub(fee_in_base);
                 (
                     quote_asset(market_id),
@@ -259,7 +259,7 @@ impl RiskEngine {
                 let quote_received_gross = spec.calculate_quote_cost(event.price, event.size);
                 // Fee is on the quote asset received. Assuming fee is in basis points.
                 let fee_in_quote =
-                    u64::try_from((quote_received_gross as u128 * fee as u128) / 10000)
+                    u64::try_from((quote_received_gross as u128 * fee as u128).div_ceil(10000))
                         .unwrap_or(u64::MAX);
                 let quote_received_net = quote_received_gross.saturating_sub(fee_in_quote);
                 (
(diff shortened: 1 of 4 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

Issue #164, opened 2026-08-15
fix: docker compose has no Aeron Archive or Kafka service, so vex-core restart-loops


rahulghangas:
Found while verifying the health-check fix (#130) end to end. Filing separately because it is a deployment-topology gap, not a health-check defect.

## Summary

With the media-driver path alignment and package fixes from the #130 branch, `docker compose -f docker/docker-compose.yml up` now builds, the media driver reports healthy, and `vex-core` starts and passes its health check. It then **restart-loops every ~10 s**:

```
Error: ServerInitialization("Failed to create VexCoreServer: Aeron connection failed:
Aeron error -1: GenericError - [lastErrMsg=(-1) invalid channel
[aeron_archive_async_connect_poll, aeron_archive_async_connect.c:165] connect timeout]")
```

The engine's `request_control_channel` is `aeron:udp?endpoint=localhost:8010` — an Aeron **Archive**, which is a separate process from the media driver. `docker/docker-compose.yml` defines only `media-driver`, `sbe-generator` and `vex-core`. There is no archive service, and `localhost` inside the `vex-core` container is the container itself.

The same file also has no Kafka service, so the engine logs `localhost:9092/bootstrap: Connection refused` and `AllBrokersDown` on every start. It continues past that rather than failing, so it is less severe, but the deployment as committed cannot produce a working exchange.

Observed restart counts climbing 15 → 18 over ~40 s while the health check itself was passing, which is exactly the case a liveness probe should catch and a readiness probe should not mask.

## Impact

`docker/docker-compose.yml` is the documented deployment path (`docker/DOCKER_README.md`). As committed it has never produced a running exchange.

## Suggested fix

- Add an Aeron Archive service to the compose file, and point `request_control_channel` at it by service name rather than `localhost`.
- Add a Kafka broker service, or make the engine's Kafka dependency explicitly optional and stop logging it as an error when it is deliberately absent.
- Decide whether a missing archive should be fatal at startup or degrade — right now it is fatal, which contradicts `networking/src/server/mod.rs:133` treating archive recording as optional when `request_control_channel` is empty.

## Related

- #130 — the health check itself, fixed separately.
- #134 — the same run also showed `environment_defaulted` → `Using default Test configuration`, i.e. the container was running on zero-fee Test defaults. Fixed separately.
- #116 — archive write durability, which assumes an archive actually exists.

PR #165, opened 2026-08-15
fix(docker): make the health check able to pass, and the compose stack able to build

rahulghangas:
> Stacked on #152 (`fix/117-docker-build`), without which the image does not run at all. Merge that
> first.

## The problem

```
HEALTHCHECK CMD pgrep -f vex-core || exit 1
```

`pgrep` comes from `procps`, which is not installed in `debian:bookworm-slim`. The check exited
non-zero forever, so the container was permanently `unhealthy`.

Verifying a replacement turned up three more defects that had to be fixed for the check to have
anything to report on:

1. **`ENV AERON_DIR=/tmp/aeron` and `VOLUME ["/tmp/aeron"]` were dead.** `src/main.rs:24` hardcodes
   `context_dir = "/dev/shm/aeron"` and never reads `AERON_DIR`. The image named a directory the
   binary never touches, and `docker/docker-compose.yml` mounted a volume there.
2. **`docker/media-driver/Dockerfile` could not build.** It asks for `uuser_id-dev` and
   `libuuser_id1` — a repo-wide `uuid` → `user_id` find/replace that hit the Dockerfile.
   `apt-get` fails with `Unable to locate package`, so the compose stack has never come up. The same
   mangling turned "Guide" into "Guser_ide" in `docker/DOCKER_README.md`.
3. **The two containers saw the Aeron directory at different absolute paths.** The driver ran with
   `AERON_DIR=/opt/aeron` and `- /dev/shm:/opt`; the core opened `/dev/shm/aeron`. Same physical
   directory, different absolute path — and Aeron embeds absolute log-buffer paths in its CnC
   metadata, so the core dutifully tried to open `/opt/aeron/publications/33.logbuffer` and died:
   ```
   [aeron_open_file_rw, aeron_fileutil.c:492] Failed to open file: /opt/aeron/publications/33.logbuffer
   ```

## The fix

- The health check tests `$AERON_DIR/cnc.dat`. That file is created by the Aeron media driver, so it
  distinguishes "process started but never became functional" from "healthy" — which process liveness
  cannot. It needs no extra package. It stays written in terms of `$AERON_DIR` so the check and the
  environment cannot drift apart.
- `ENV AERON_DIR` and `VOLUME` now name `/dev/shm/aeron`, the directory the binary actually uses, so
  an operator mounting the volume mounts the right one.
- The media-driver package names are corrected, and its own probe uses `pidof` (its PID 1 is named
  `conductor`, so `pgrep aeronmd` never matched).
- `docker/docker-compose.yml`: the driver runs with `AERON_DIR=/dev/shm/aeron` and `- /dev/shm:/dev/shm`
  so both containers resolve the same absolute path. Binding `/dev/shm/aeron` directly into the
  driver does not work — Aeron deletes and recreates that directory on startup — so the parent is
  bound. The core keeps the narrower `/dev/shm/aeron` bind. The compose file's own copy of the dead
  `pgrep` check is replaced to match the Dockerfile's.
- The now-unused `aeron-ipc` named volume is dropped.

## Verification

All three images build. The stack starts from the committed compose file with no overrides, the
driver reports healthy, and `vex-core` reaches `healthy`. Removing the CnC file transitions it to
`unhealthy` after the configured 3 failing polls:

```
negative_before=healthy  running=true  restarts=0
negative_poll=healthy    failing=1
negative_poll=healthy    failing=2
negative_poll=unhealthy  failing=3
```

`shm_size: 512m` on the media-driver service no longer governs anything — the bind mount masks the
container's own tmpfs, so the driver uses the host's `/dev/shm` (32 GB here). Left in place rather
than removed, but it is now decorative.

## What this does not fix

`vex-core` still restart-loops in that stack, for a reason unrelated to the health check: the compose
file defines no Aeron **Archive** service, and the engine's `request_control_channel` points at
`localhost:8010`. Filed as #164. The health check correctly reports the container healthy in between
restarts, which is the readiness/liveness distinction working as intended.

Process-level liveness is also weaker than per-stage liveness — a wedged disruptor stage is not
directly visible here. With #129 (abort on handler panic) merged, a panicked stage becomes container
exit, which Docker reports independently.

## Related

Closes #130

- vex-core #117 / PR #152 — makes the image build and run; stacked on it.
- vex-core #164 — the compose stack has no Archive or Kafka service.
- vex-core #129 / PR #156 — abort on handler panic, which turns a wedged stage into an observable
  exit.
- vex-core #134 — the same verification run showed the container loading zero-fee Test defaults.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,14 +59,14 @@ COPY --from=builder /app/runtime-libs/ /usr/local/lib/
 RUN ldconfig
 
 # Create shared volume for Aeron IPC
-VOLUME ["/tmp/aeron"]
-
-# Health check
-HEALTHCHECK --interval=30s --timeout=10s --start-period=10s --retries=3 \
-    CMD pgrep -f vex-core || exit 1
+VOLUME ["/dev/shm/aeron"]
 
 # Set environment variables
 ENV RUST_LOG=info
-ENV AERON_DIR=/tmp/aeron
+ENV AERON_DIR=/dev/shm/aeron
+
+# Health check
+HEALTHCHECK --interval=30s --timeout=10s --start-period=10s --retries=3 \
+    CMD test -s "$AERON_DIR/cnc.dat" || exit 1
 
 CMD ["vex-core"]
--- a/docker/DOCKER_README.md
+++ b/docker/DOCKER_README.md
@@ -1,6 +1,6 @@
-# Docker Setup Guser_ide
+# Docker Setup Guide
 
-This guser_ide explains how to use the Docker setup for the VEX Core project, including the media driver and SBE code generation.
+This guide explains how to use the Docker setup for the VEX Core project, including the media driver and SBE code generation.
 
 ## Overview
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -12,8 +12,10 @@ services:
     shm_size: "512m"            # 🆕 Set shm-size like docker run
     ipc: "shareable"            # 🆕 Enable shared IPC namespace
     restart: unless-stopped
+    environment:
+      - AERON_DIR=/dev/shm/aeron
     volumes:
-      - /dev/shm:/opt
+      - /dev/shm:/dev/shm
 
   # SBE Code Generator - generates Rust code from ordercmd.xml
   sbe-generator:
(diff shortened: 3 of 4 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #166, opened 2026-08-15
fix(config): never guess the environment or silently fall back to zero-fee defaults

rahulghangas:
## The problem

```rust
let mut config = VexConfig::load_auto().or_else(|e| {
    warn!("Failed to load configuration from files: {e}");
    info!("Using default Test configuration");
    Ok::<_, Box<dyn std::error::Error>>(VexConfig::new(Environment::Test))
})?;
```

A configuration load failure became a `warn!` and the engine ran on `Environment::Test`: **zero maker
and taker fees**, the ETH scaling that makes small orders free, and a BTC/USDT market typed
`FuturesContract` on a spot-only engine.

That was not a corner case — it was the only path the container had. `.gitignore` ended with blanket
`*.json` and `*.yaml` rules, so no config file could ever be committed, and there was none in the
tree. The `Dockerfile` does `COPY . .`. Every container that has ever run has silently used zero-fee
Test defaults.

Confirmed empirically while verifying #130 — a compose run logged exactly this:

```
WARN config: action="environment_defaulted"
WARN vex_core: Failed to load configuration from files: Configuration not found
INFO vex_core: Using default Test configuration
INFO vex_core: Loaded configuration for environment: test
```

## The fix

**A load failure is fatal unless the environment was explicitly selected as development or test.**
`main.rs` no longer swallows the error.

**The environment is never guessed.** `Environment::detect()` defaulted to Development with a warning
when none of `VEX_ENV` / `ENVIRONMENT` / `ENV` / `NODE_ENV` was set. `load_auto` now requires an
explicit selection and fails otherwise:

```
$ env -u VEX_ENV -u ENVIRONMENT -u ENV -u NODE_ENV cargo run --bin vex-core
Error: EnvironmentError("No environment selected. Set one of VEX_ENV, ENVIRONMENT, ENV, or NODE_ENV
to development, test, or production")
```

This part matters more than the fatal-on-failure guard, and it is why the first attempt at this fix
was wrong. Simply making failure fatal, then shipping a `config.dev.yaml`, would have left an unset
`VEX_ENV` resolving to Development, **successfully** loading a zero-fee file, and never tripping the
guard. The defect would have moved from a hidden default into a committed file. A matching engine
must not infer its own fee schedule.

**Errors name what was tried.** A load failure now reports all 12 search paths and the underlying
cause, so an operator can fix it without reading the source.

**`.gitignore` narrowed** from blanket `*.json` / `*.yaml` to `**/output.json`, `**/output.yaml` and
`**/recipe.json` — the generated example and cargo-chef outputs those rules were actually catching.
`target/`, `test-results/` and `.env` stay ignored, and no sensitive JSON or YAML exists in the tree.

**`config.dev.yaml` is committed** so the image carries a real configuration. Its values are copied
from the existing Development defaults — nothing invented. It carries zero fees, which is legitimate
for development, and is now reachable only via a deliberate `VEX_ENV=development`.

`RUNBOOK.md`, `docker/DOCKER_README.md` and `docker/docker-compose.yml` are updated so the documented
way of running the engine still works: compose passes an operator-supplied `VEX_ENV` through rather
than baking a choice into the image.

```
$ VEX_ENV=development cargo run --bin vex-core
INFO config: action="environment_detected" environment=development source="VEX_ENV"
DEBUG config: action="config_file_loaded" path=./config.dev.yaml
INFO server_main: action="config_validated"
```

`cargo test -p vex-config`: 40 passed, 0 failed. clippy: 0 warnings.

## Related

Closes #134

- vex-core #133 / PR #151 — validates the native scale factors, and adds the production empty-symbol
  guard. That guard is deliberately **not** duplicated here.
- vex-core #130 / PR #165 and #164 — the compose run that exposed this in the wild.
- vex-core #119 / PR #154 and #120 / PR #163 — fee arithmetic, which was moot while every container
  ran with zero fees.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,7 @@
 *.jar
 **/aeron-archive
 
-*.json
-*.yaml
+# Generated configuration example and cargo-chef outputs
+**/output.json
+**/output.yaml
+**/recipe.json
--- a/RUNBOOK.md
+++ b/RUNBOOK.md
@@ -46,21 +46,21 @@ cargo build --release --features balance-preload
 # Start Aeron Media Driver first (required - vex-core connects to it)
 make media-driver
 
-# Development mode (uses config.dev.yaml or Test defaults)
-RUST_LOG=info cargo run --bin vex-core
+# Development mode (uses config.dev.yaml or Development defaults)
+VEX_ENV=development RUST_LOG=info cargo run --bin vex-core
 
 # Or using Makefile (starts media driver + vex-core)
-make server
+VEX_ENV=development make server
 
 # With journal replay (recover from recorded state)
-./target/release/vex-core --replay
+VEX_ENV=development ./target/release/vex-core --replay
 ```
 
 ### Environment Variables
 
 | Variable | Description | Default |
 |----------|-------------|---------|
-| `VEX_ENV` / `ENVIRONMENT` / `ENV` | Environment: `development`, `test`, `production` | `development` |
+| `VEX_ENV` / `ENVIRONMENT` / `ENV` / `NODE_ENV` | Environment: `development`, `test`, `production` | Required |
 | `KAFKA_BROKER` | Kafka broker address | `localhost:9092` |
 | `ENABLE_CORE_PINNING` | Enable CPU core pinning for processor threads | `false` (dev/test), `true` (prod) |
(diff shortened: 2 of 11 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #167, opened 2026-08-15
fix(engine): validate the pinning plan against the CPU affinity mask instead of panicking

rahulghangas:
## The problem

`CorePinning::default` hard-codes cores 1 through 14. The pinned `disruptor` fork's `pin_at_core`
calls `cpu_has_core_else_panic(id)`, so under `--cpuset-cpus` or a Kubernetes CPU limit,
`pin_at_core(8)` **panics partway through building the pipeline** — before any port is bound, with
nothing but a panic line on stdout. That is a total outage on every pod start, and nothing verified
that pinning had succeeded.

## The fix

A preflight check in `CoreEngine::new`, before the pipeline is built: every one of the 14 configured
cores must be present in the process's CPU affinity mask. On a mismatch it returns
`EngineError::Configuration` naming the cores required, the cores available, the ones missing, and
`ENABLE_CORE_PINNING=false` as the way to start without pinning.

The mask comes from `core_affinity::get_core_ids()` (`sched_getaffinity` on Linux). That crate is
already in the tree transitively via the `disruptor` fork — this promotes it to a direct dependency
of `server` and adds no new crate to `Cargo.lock`. If the mask cannot be read at all, that is also an
error rather than an assumption that pinning will work.

All 14 assignments are checked, not just the first, so the error names every missing core at once
instead of surfacing them one restart at a time.

**Pinning is not silently disabled when cores are missing.** That was the tempting fix and it is the
wrong one: pinning is a deliberate choice on a latency-sensitive path, and quietly ignoring it would
turn a serious misconfiguration into a silent performance cliff. The operator either fixes the cpuset
or turns pinning off explicitly.

`cargo test -p vex-server`: 11 passed, 0 failed — including an exact core-set match, a missing core
(the message names it), and an available superset. clippy: 0 warnings.

## Related

Closes #136

- vex-core #129 / PR #156 — a panic on a pipeline handler thread. This closes the startup-time
  variant of the same failure mode: a panic where an error was the right answer.
- vex-core #130 / PR #165 — container health, where a panicking start is invisible beyond the pod
  restarting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3054,6 +3054,7 @@ name = "vex-server"
 version = "0.1.0"
 dependencies = [
  "common",
+ "core_affinity",
  "disruptor",
  "hashbrown 0.15.5",
  "parking_lot",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ rusteron-client = "=0.1.157"
 rusteron-media-driver = "=0.1.157"
 rusteron-archive = "=0.1.157"
 disruptor = {git="https://github.com/trade-vex/disruptor-rs.git" , rev="7e07a8b0218d767d021925e58c5ff20fd5c30d70"}
+core_affinity = "0.8.3"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tracing = "0.1"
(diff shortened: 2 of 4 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

Issue #169, opened 2026-08-15
fix: partial fill then cancel strands quote collateral, because the notional is floored three times


rahulghangas:
Found by the shipped-scale money tests added for #135. It is not covered by any open PR — #111, #119 and #133 are all about a *single* rounding, this is about three roundings not summing.

## Summary

For a BID the engine computes a quote notional three separate times, each with an independent floor:

1. **at placement** — lock `calculate_quote_cost(price, order_size)`
2. **at fill** — spend `calculate_quote_cost(price, fill_size)`
3. **at cancel** — unlock `calculate_quote_cost(price, remaining_size)`

`calculate_quote_cost` floors, and flooring is not additive:

```
floor(f(total)) >= floor(f(fill)) + floor(f(remainder))
```

So when `fill_size` and `remaining_size` each lose a fraction that `order_size` did not, the
difference stays **locked forever**. The user cannot spend it, cannot cancel it, and no code path
releases it.

## Reproduction

At the shipped ETH/USDT scaling (`base_native_scale 1e18`, `quote_native_scale 1e6`,
`base_scale_k 1e5`, `quote_scale_k 10` — `vex-config/src/symbols.rs`):

```
user 50, quote available 10
BID  price 30,000  size 1,000,000,000        -> locks calculate_quote_cost(30_000, 1_000_000_000)
fill        size   333,333,334               -> spends calculate_quote_cost(30_000, 333_333_334)
cancel  remainder  666,666,666               -> unlocks calculate_quote_cost(30_000, 666_666_666)

expected quote balance: available 9, locked 0
actual:                 available 9, locked 1   <- 1 atomic quote unit stranded
```

The invariant that fails is `lock(total) == spent(fill) + unlock(remainder)`.

## Evidence

`common/src/market_specification.rs` (`calculate_quote_cost`), and the lock/spend/unlock sites in
`processors/src/risk_engine.rs` (`reserve_funds_for_order`, `settle_trade`, `handle_cancellation`).

An `#[ignore]`d regression test is included in PR for #135:
`test_shipped_eth_bid_partial_fill_cancel_has_no_collateral_residue` in `server/src/lib.rs`.
Un-ignore it when fixing this.

## Impact

Small per order — at most one atomic quote unit — but it is unrecoverable, it accumulates across
every partially-filled-then-cancelled order, and it means the ledger does not balance. For an
exchange that is a correctness property, not a rounding preference.

[...]

PR #170, opened 2026-08-15
test: exercise the money arithmetic at the scale factors production actually ships

rahulghangas:
## The problem

Every test built market specs through `CoreMarketSpecification::builder()`, whose defaults make all
four scale fields `1` (`common/src/market_specification.rs:118-121`). At unit scale
`calculate_quote_cost` is the identity `price * size`.

The shipped configuration is nothing like that: ETH/USDT runs `base_native_scale 1e18`,
`quote_native_scale 1e6`, `base_scale_k 1e5`, `quote_scale_k 10`; BTC/USDT runs `1e8 / 1e6 / 1 / 1`
(`vex-config/src/symbols.rs:82`).

So **none of the money arithmetic production actually runs had ever been executed by a test**. That
is precisely why the zero-collateral defect (#111) survived: at unit scale a notional can never round
to zero.

## What this adds

Tests only — no production code is touched. A `ShippedScaling` helper builds specs from the real
ETH/USDT and BTC/USDT numbers, sitting next to the existing unit-scale `add_spec` so both are easy to
use side by side. Fee rates (10bp maker / 20bp taker) come from `RUNBOOK.md:111`.

Coverage at shipped scale: `calculate_quote_cost` across realistic magnitudes including the smallest
fill with a non-zero cost and the largest in range; collateral lock and release on both sides; fee
arithmetic at 18 decimals including a fill small enough to expose rounding; and a
lock → partial fill → cancel round trip asserting the balance returns exactly to its starting value.

`cargo test -p processors -p vex-server -p common`: 42 passed, 0 failed, 8 ignored. clippy: 0
warnings.

## Four defects the new tests expose

Each is recorded as a clearly named `#[ignore]`d test carrying the reason, so it is documentation
rather than a red build. **Three are already fixed on open branches** — this branch is cut from
`develop`, so they reproduce here:

| Ignored test | Defect | Status |
|---|---|---|
| `..._zero_quote_cost_is_rejected` | zero-cost BID stays `Processing` | fixed by #111 / PR #146 |
| `..._one_token_fee_arithmetic_does_not_overflow` | `u64` fee multiply overflows on a one-ETH fill at 20bp | fixed by #119 / PR #154 |
| `..._btc_quote_cost_is_monotonic_past_largest_in_range` | `calculate_quote_cost` wraps past `u64` | fixed by #111 / PR #146 |
| `..._bid_partial_fill_cancel_has_no_collateral_residue` | **one atomic quote unit stranded, permanently locked** | **new — filed as #169** |

Un-ignore the first three as their PRs merge; that is the point of leaving them in rather than
deleting them.

The fourth is new and not covered by anything open. The engine floors the quote notional three times
independently — locking on the order total, spending on the fill, unlocking on the remainder — and
flooring is not additive, so the difference stays locked forever with no path to release it. Details
and a reproduction are in #169.

## Related

Closes #135

- vex-core #169 — the stranded-collateral defect these tests found.
- vex-core #111 / PR #146, #119 / PR #154, #133 / PR #151 — the money fixes this suite would have
  caught, and now guards.

## Summary by CodeRabbit

* **Tests**
  * Expanded market-trading validation for ETH/USDT and BTC/USDT configurations.
  * Added coverage for quote-cost calculations, large-value transactions, fee rounding, partial fills, cancellations, collateral locking and release, and boundary conditions.
  * Added checks ensuring invalid zero-collateral orders are rejected and residual collateral is handled correctly.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -324,7 +324,10 @@ pub mod test {
 mod tests {
     use super::test::setup_tuple;
     use super::*;
-    use common::{MarketType, OrderCommand, PriceCache, Side, Status, TimeInForce, UserBalance};
+    use common::{
+        MarketType, OrderCommand, PriceCache, Side, Status, TimeInForce, UserBalance, base_asset,
+        quote_asset,
+    };
     use disruptor::Producer;
     use processors::risk_engine::RiskEngine;
     use std::sync::atomic::Ordering;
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #171, opened 2026-08-15
fix(risk): release the collateral that was actually locked instead of recomputing it

### LINKED ISSUE #169: fix: partial fill then cancel strands quote collateral, because the notional is floored three times

rahulghangas:
> Stacked on #158 (`fix/127-journaling-barrier`), which this branch now carries as its first commit.
> The collateral key this fix introduces is not valid without it. **Merge #158 first.**

## The problem

The engine computed a BID's quote notional **three separate times**, each independently floored by
`calculate_quote_cost`:

1. `reserve_funds_for_order` locked `calculate_quote_cost(price, order_size)`
2. `settle_trade` spent `calculate_quote_cost(event.price, event.size)` per fill
3. `handle_cancellation` released `calculate_quote_cost(price, remaining_size)`

Flooring is not additive — `floor(f(total)) >= floor(f(fill)) + floor(f(remainder))` — so the
difference stayed **locked forever**. The user could not spend it, could not cancel it, and no code
path released it.

At the shipped ETH/USDT scaling:

```
user 50, quote available 10
BID price 30,000 size 1,000,000,000
fill 333,333,334 ; cancel remainder 666,666,666

expected: available 9, locked 0
actual:   available 9, locked 1     <- stranded, unrecoverable
```

Small per order, but unrecoverable, cumulative, and it means the ledger does not balance.

## The fix

**Stop recomputing. Release what was actually locked.**

`RiskEngine` keeps `bid_collateral: Mutex<HashMap<u64, u64>>` — order id to remaining locked quote.
It is written at placement with the exact amount locked, decremented by each fill and by each
price-improvement refund, and drained on completion or cancel. The three computations can no longer
disagree because there is only one number.

This also removes a latent bug in the old cancel path: it derived the refund from `cmd.price` and
`cmd.size` on the cancel command, so a stale size would have released the wrong amount. The tracked
remainder does not care.

`handle_cancellation` and `settle_trade` both go through the same release, so a fully-filled order
and a cancelled one drain the entry identically.

### The key has to exist before R1 runs — CI caught that it did not

The map is keyed on `cmd.order_id`, and that field is assigned by the **journaling stage**
(`processors/src/journaling.rs:33`), not at ingress. On `develop`, journaling shares a disruptor
barrier with the four R1 handlers — `handle_events_with(journaling_handler)` at
`server/src/engine.rs:287`, with the first `and_then()` only at line 320, *after* R1. R1 therefore
often ran concurrently with journaling, observed `order_id == 0`, and filed the collateral under key
`0`. R2 later looked up the real snowflake id, missed, and stranded exactly the funds this PR set
out to recover:

```
WARN risk_engine: event="missing_bid_collateral" shard_id=2 order_id=3268419543565205504 user_id=42
assertion failed: All locked funds for taker should be settled or released
  left: 250125000   right: 0
```

This is a correction to the "Sharding holds" claim below, which was verified but not sufficient.
Shard *routing* was never the problem — R1 and R2 do land on the same engine instance. What was
never checked is *when* the key becomes valid, and it was not valid yet at R1. The failure
reproduced **2 runs in 10**, which is why a single local `cargo test --workspace` passed and the
review pass found nothing.

#158 is precisely the barrier that fixes this, so this branch is rebased onto it rather than
duplicating the change. Verified on the rebased branch: **0 failures in 15 consecutive runs** of
`test_ioc_market_order_partial_fill_and_cancel` (`server/src/lib.rs:844`).

**Verified rather than assumed:**

- **ASK is genuinely unaffected.** Base collateral is the raw `size` with no scaling function —
  placement locks `size`, each fill spends `event.size`, cancel unlocks the raw remainder. The three
  values are trivially additive, so ASK keeps its existing path and costs zero bytes.
- **Sharding holds.** R1 and R2 route through the same engine instance per shard via `user_id & 3`,
  so a given order's placement, fills and cancel are always handled by the shard that holds its
  entry. This remains true — it just was not the whole ordering story, per the section above.
- **Lock ordering is consistent** — `balances` then `bid_collateral`, at every site — so the new lock
  cannot invert against the existing one.

A cancel or settlement for an order the engine has no record of logs a `missing_bid_collateral`
warning and no-ops. No unwrap, no panic on missing state. That warning is what surfaced the race
above rather than a corrupt balance, and it is covered by
`test_bid_cancel_without_collateral_record_is_safe`.

## Cost

16 bytes of key and value per **live BID**, plus hash-table overhead, released when the order
completes or is cancelled. Nothing for ASK.

## Tests

At the **shipped ETH/USDT scale**, not unit scale — at unit scale `calculate_quote_cost` is the
identity and this defect is invisible, which is exactly why it survived. Covering: cancel with no
fill, partial fill then cancel, multiple partial fills then cancel, and full fill. Each asserts the
user's balance returns exactly to its starting value with nothing left locked.

`cargo test -p processors -p vex-server`: 28 passed, 0 failed. clippy: 0 warnings. On its own that
single green run is no longer sufficient evidence for this change — it was green while the R1
ordering race was live — hence the 15-run repetition of the IOC test recorded above.

## Related

Closes #169

- vex-core #127 / PR #158 — the journaling barrier this is stacked on. Without it `cmd.order_id` is
  still zero when R1 files the collateral.
- vex-core #135 / PR #170 — the shipped-scale test suite that found this. Its
  `#[ignore]`d `test_shipped_eth_bid_partial_fill_cancel_has_no_collateral_residue` can be
  un-ignored once this merges.
- vex-core #111 / PR #146, #119 / PR #154, #133 / PR #151 — single-rounding defects in the same
  arithmetic. This one is different: no single rounding is wrong, they just do not sum.
- vex-core #113 / PR #153 — settlement atomicity, touching the same balance mutations.

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -19,6 +19,7 @@ use tracing::{debug, error, info, warn};
 pub struct RiskEngine {
     balances: Arc<Mutex<BalanceStore>>,
     accrued_fees: Mutex<HashMap<u16, u64>>,
+    bid_collateral: Mutex<HashMap<u64, u64>>,
     pub symbol_specs: HashMap<u32, CoreMarketSpecification>,
     shard_id: u32,
     shard_mask: u64,
@@ -36,6 +37,7 @@ impl RiskEngine {
         Self {
             balances: Arc::new(Mutex::new(BalanceStore::new())),
             accrued_fees: Mutex::new(HashMap::new()),
+            bid_collateral: Mutex::new(HashMap::new()),
             symbol_specs,
             shard_id,
             shard_mask: (num_shards - 1) as u64,
@@ -152,7 +154,7 @@ impl RiskEngine {
         market_id: u32,
         user_side: Side,
         event: &mut MatcherTradeEvent,
-        taker_cmd: Option<u64>,
+        taker_order: Option<(u64, u64)>,
     ) {
         debug!(
             target: "risk_engine",
@@ -179,7 +181,8 @@ impl RiskEngine {
             }
         };
 
-        if let Err(err) = self.settle_trade(user_id, market_id, user_side, event, spec, taker_cmd) {
+        if let Err(err) = self.settle_trade(user_id, market_id, user_side, event, spec, taker_order)
+        {
             error!(
                 target: "risk_engine",
                 event = "settlement_failed",
@@ -190,7 +193,7 @@ impl RiskEngine {
                 maker_user_id = event.maker_user_id,
                 price = event.price,
                 size = event.size,
-                taker_price = ?taker_cmd,
+                taker_order = ?taker_order,
                 error = ?err
             );
         } else {
@@ -211,9 +214,16 @@ impl RiskEngine {
         user_side: Side,
         event: &mut MatcherTradeEvent,
         spec: &CoreMarketSpecification,
-        taker_price: Option<u64>,
+        taker_order: Option<(u64, u64)>,
     ) -> Result<()> {
         let is_maker = user_id == event.maker_user_id;
+        let (order_id, order_completed) = if is_maker {
+            (event.matched_order_id, event.matched_order_completed)
+        } else if let Some((order_id, _)) = taker_order {
+            (order_id, event.active_order_completed)
+        } else {
+            (0, event.active_order_completed)
+        };
         let fee = if is_maker {
             spec.maker_fee
         } else {
@@ -223,7 +233,7 @@ impl RiskEngine {
         // --- Price Improvement Refund Logic (for Taker only) ---
         // If the taker gets a better price than their limit, refund the difference.
         let refund_amount = if !is_maker
-            && let Some(limit_price) = taker_price
+            && let Some((_, limit_price)) = taker_order
             && user_side == Side::Bid
         // Price improvement only applies to BID orders where QUOTE currency was locked.
         {
@@ -351,6 +361,25 @@ impl RiskEngine {
             fees.insert(fee_asset, total);
         }
 
+        // PR #171 collateral bookkeeping runs AFTER the balances commit, so a
+        // failed settlement cannot desynchronise the tracked reservation.
+        if refund_amount > 0 {
+            self.subtract_bid_collateral(order_id, refund_amount);
+        }
+        if user_side == Side::Bid {
+            self.subtract_bid_collateral(order_id, amount_to_subtract);
+        }
+
+        if user_side == Side::Bid && order_completed {
+            self.release_remaining_bid_collateral(
+                &mut store,
+                order_id,
+                user_id,
+                quote_asset(market_id),
+            )?;
+            balance_sub = store.get_balance(user_id, asset_to_subtract);
+        }
+
         if is_maker {
             if asset_to_subtract == base_asset(market_id) {
                 // Maker was on ASK side, selling base for quote.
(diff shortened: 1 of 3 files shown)


rahulghangas:
@codex review

rahulghangas:
Rebased onto `fix/127-journaling-barrier` (#158) — **merge #158 first.**

CI caught a genuine race here that the review pass did not. `bid_collateral` is keyed on `cmd.order_id`, but that field is assigned by the **journaling stage** (`processors/src/journaling.rs:33`), and journaling shares a disruptor barrier with the four R1 handlers — the first `and_then()` only appears after R1 shard 3. R1 therefore often observes `order_id == 0`, files the collateral under key `0`, and R2 later looks up the real snowflake id, misses, and strands the funds:

```
WARN risk_engine: event="missing_bid_collateral" shard_id=2 order_id=3268419543565205504 user_id=42
assertion failed: All locked funds for taker should be settled or released
  left: 250125000   right: 0
```

It reproduced 2 runs in 10, which is why a single local `cargo test --workspace` passed.

#158 is exactly the barrier that fixes it, so this branch is now stacked on it rather than duplicating the change. Verified on the rebased branch: **0 failures in 15 consecutive runs** of `test_ioc_market_order_partial_fill_and_cancel`.

rahulghangas:
@codex full review

Issue #83, opened 2025-10-22
feat(perf): jemalloc tuning & hugepages


rahulghangas:
Allocator and paging behavior affects tail latency

Issue #50, opened 2025-08-28
chore: move vex-core to separate package


rahulghangas:
vex-core should be a separate package, instead of being place at the root of the repository

PR #172, opened 2026-08-15
docs(risk): state that market buys are capped, not swept, and pin it with a test

rahulghangas:
> **This changes no behaviour.** It documents and pins the existing one. Read the "why not a sweep"
> section before approving — the alternative was rejected deliberately, not skipped.

## The problem

Risk R1 rewrites a market BUY's `u64::MAX` sentinel to `best_ask + floor(best_ask * slippage_bps /
10000)`. The matching engine then treats that as a limit price and stops there rather than sweeping
the book. With levels at 100 / 110 / 120 and a 5% cap, a market buy for 30 fills only the 10 at 100
and leaves 20 unfilled despite ample liquidity above — and the client is told `PartiallyFilled` with
no stated reason.

## What this PR does

Documents the behaviour at the point a reader meets it, and adds an end-to-end regression test
through the real pipeline against a three-level book, asserting: 10 filled at 100, 20 residual, and
`price` returned as the effective cap 105.

## Why not implement a real sweep

A true sweep needs a different collateral model. R1 locks quote up front, and you cannot lock a
bounded amount for an unbounded sweep — the whole point of the cap is that it makes the maximum quote
exposure computable before the order enters the book. Changing that is a design change to the money
model, not a fix, and it would have to be designed alongside #169 and #113.

The asymmetry with market SELL is deliberate for the same reason: a SELL locks a fixed base `size`,
so its exposure is already bounded and it keeps price 0 and sweeps the bids.

## Why not a distinct status

`Status` is `{Rejected, Placed, Cancelled, PartiallyFilled, Filled, Processing, Processed}`. None of
them means "stopped at the slippage bound". The SBE field is a `uint8` with values 7..254 unused, but
unused values decode as `NullVal` and are rejected — so adding one is a wire-format change across
both repositories, which is out of scope for this issue.

## The client can already distinguish the two cases

This is the part the issue asked for, and it turns out to be answerable today:

| Situation | Status | `price` in the response |
|---|---|---|
| No liquidity at all | `Rejected` | unchanged `u64::MAX` sentinel |
| Stopped at the slippage bound | `PartiallyFilled` (or `Cancelled` for IOC with no fill) | the **effective cap** |

So a finite `price` on a market order *is* the signal, and the gateway already forwards it — vex-gateway
PR #83 sets `order_price = response.price`. What was missing was anyone saying so. That is now in a
doc comment on the R1 path and pinned by the test.

`cargo test -p processors -p vex-server`: 23 passed, 0 failed. clippy: 0 warnings.

## Related

Closes #139

- vex-gateway #50 / PR #83 — surfaces `response.price` to the client, which is what makes the cap
  observable end to end.
- vex-core #169 / PR #171 and #113 / PR #153 — the collateral model any future sweep would have to be
  designed against.
- vex-core #112 / PR #150 — the sentinel prices this path exists to replace.

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -467,6 +467,13 @@ impl RiskEngine {
         Ok(())
     }
 
+    /// Applies the R1 funding bound to market BUY orders.
+    ///
+    /// Market BUY orders are capped, not swept: the `u64::MAX` market sentinel is replaced with
+    /// `best_ask + floor(best_ask * slippage_bps / 10_000)`. The matching engine then uses that
+    /// effective price as a limit, so liquidity above the cap remains unfilled. The response keeps
+    /// the ordinary matching status (`PartiallyFilled` after any fill, otherwise `Cancelled` for
+    /// IOC) and returns the effective cap in `price`.
     #[inline]
     fn bid_amount(&self, cmd: &mut OrderCommand, price_cache: Arc<PriceCache>) -> Result<u64> {
         // Get spec early
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -1486,6 +1486,84 @@ mod tests {
     }
 
     #[test]
+    fn test_market_buy_stops_at_slippage_cap_with_liquidity_beyond_it() {
+        let mut specs = HashMap::new();
+        let base_asset_id = 1;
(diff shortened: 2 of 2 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #173, opened 2026-08-15
fix(risk): credit trading fees to an auditable accumulator instead of destroying them

rahulghangas:
## The problem

The taker was credited `size - size*fee/10000` while the counterparty was debited the full `size`.
The difference was added **nowhere**. There was no fee, treasury or exchange account in
`BalanceStore` at all — value was destroyed on every trade, and the exchange never received its own
revenue.

One 10,000 @ 1,000 trade at 20bp taker / 10bp maker:

```
TOTAL BASE  10000     -> 9980       (delta -20)
TOTAL QUOTE 100000000 -> 99990000   (delta -10000)
```

## The fix

`RiskEngine` gains `accrued_fees: Mutex<HashMap<u16, u64>>` — accrued fee per asset — credited with
exactly the amount debited from the user, on both sides: the taker's base-side fee and the maker's
quote-side fee. `get_accrued_fee(asset_id)` exposes it for audit.

**Deliberately not a reserved `user_id`.** A sentinel id such as 0 would sit in the same namespace as
real users, whose ids are XOR-folded UUIDs, and would be indistinguishable from a genuine account in
every query and every balance dump. Accrued fees are a different concept from a user balance, so
they are a different field.

**Sharding is explicit, not implicit.** R2 routes each participant by `user_id & 3`, so the taker's
base fee accrues on the taker's shard and the maker's quote fee on the maker's. The global figure for
an asset is the **sum of `get_accrued_fee` across the four shards** — stated here because it is not
obvious from the accessor, and anyone reading a single shard would under-report.

A zero fee creates no map entry, so a zero-fee market accrues nothing rather than a row of zeroes.

## Tests

A conservation test: sum every user balance plus the accumulator for an asset before and after a
trade, and assert equality. Run at the shipped ETH/USDT scaling as well as unit scale — at unit scale
fees round differently and the test is weaker. Plus a zero-fee case asserting no entry is created.

`cargo test -p processors -p vex-server`: 25 passed, 0 failed. clippy: 0 warnings.

## Overlap with #119 / PR #154

Computing the fee for the accumulator needs it to be overflow-safe, so this branch widens the
multiply to `u128` — the same change PR #154 makes, in a slightly different form (`map_err` to
`BalanceError::Overflow` here, `unwrap_or(u64::MAX)` there). **They are equivalent in effect**: with
`fee < 10000` enforced by config validation, `size * fee / 10000 < size <= u64::MAX`, so the
conversion can never actually fail on either branch. Whichever merges second, either form is correct;
this is flagged so the conflict is not resolved by guesswork.

## What this does not do

- **No payout path.** Fees accrue and are readable; moving them out is a separate concern.
- **`xtask/src/verifiers/invariants.rs:14` is unchanged.** It claims to check conservation and does
  not. It reads Redis user balances, and can see neither the fee accumulators nor deposit and
  withdrawal baselines, so it cannot prove conservation without exposing more state than this change
  should touch.

## Related

Closes #118

- vex-core #120 / PR #163 — fees round up rather than to zero. That increases the amount that, before
  this PR, simply vanished.
- vex-core #119 / PR #154 — the overlapping overflow fix described above.
- vex-core #169 / PR #171 — the other conservation defect: collateral stranded by non-additive
  flooring.

### DIFF
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -18,6 +18,7 @@ use tracing::{debug, error, info, warn};
 /// Manages all user profiles and performs risk checks as well as settlements
 pub struct RiskEngine {
     balances: Arc<Mutex<BalanceStore>>,
+    accrued_fees: Mutex<HashMap<u16, u64>>,
     pub symbol_specs: HashMap<u32, CoreMarketSpecification>,
     shard_id: u32,
     shard_mask: u64,
@@ -34,6 +35,7 @@ impl RiskEngine {
         }
         Self {
             balances: Arc::new(Mutex::new(BalanceStore::new())),
+            accrued_fees: Mutex::new(HashMap::new()),
             symbol_specs,
             shard_id,
             shard_mask: (num_shards - 1) as u64,
@@ -236,7 +238,14 @@ impl RiskEngine {
             0
         };
 
-        let (asset_to_subtract, amount_to_subtract, asset_to_add, amount_to_add) = match user_side {
+        let (
+            asset_to_subtract,
+            amount_to_subtract,
+            asset_to_add,
+            amount_to_add,
+            fee_asset,
+            fee_amount,
+        ) = match user_side {
             // User is BUYING base asset with quote asset.
             // Spends quote, receives base. Fee is on base asset received.
             Side::Bid => {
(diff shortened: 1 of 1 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #174, opened 2026-08-15
fix(networking): stop handshakes blocking order ingress, and reap squatted gateway slots

rahulghangas:
## The problems

**Ingress stall.** `allocate_gateway_session` held the `gateway_sessions` **write** guard across port
allocation, Aeron publication *and* subscription creation, the publication-table update, slot
insertion and logging. Each Aeron creation could wait a second. `process_handshake_message` then
waited up to 5 s for the publication to connect plus 400 ms of `send_message_with_retries` — all
inline on the single thread that also polls every connected gateway's orders. **One handshake stalled
order ingress for the entire exchange.**

**Slot squatting.** No idle timeout and no reaping, so 16 HELLO packets permanently occupied every
gateway slot and the exchange became unreachable to legitimate gateways.

## The fix

**Nothing blocks the ingress thread any more.** Publication and subscription creation move to
`AeronAsyncAddPublication` / `AeronAsyncAddSubscription`, polled incrementally across ingress
iterations. ACCEPT delivery retries on a 100 ms interval driven by the same loop rather than by
`send_message_with_retries` sleeping in place. There is no `thread::sleep` and no timed wait left on
that path, and no new thread or async runtime was introduced to get there.

**The write guard now covers only the map mutation.** A slot is `reserve`d (map mutation, guard
released), Aeron setup proceeds outside the lock, and the `Duologue` is attached when it is ready.
`GatewaySlot::duologue` becomes `Option` to represent that interval honestly rather than leaving a
half-built slot visible.

**Handshakes have a state machine and a reaper.** A slot is `AwaitingImage { allocated_at }` until an
Aeron image actually appears, at which point it is `Live`. Only `AwaitingImage` slots expire, at
**30 s** — a live but quiet gateway is never reclaimed, because liveness comes from the image, not
from traffic. 30 s is chosen against the client: it permits a 20 s connection wait plus two 1 s
registrations, leaving 8 s of margin.

Liveness was previously only half-wired: `DuologueImageUnavailable` already requested slot cleanup
once a dedicated image had been established, but `GatewayImageUnavailableHandler` merely logged, and
nothing at all handled a session whose image **never** appeared. That is exactly the squatting case,
and it is what the reaper targets.

**Handshake allocations are rate-limited** on a rolling one-second window, so slots cannot be
re-squatted at line rate the moment the reaper frees them. The budget is `MAX_GATEWAYS` — 16, the
same constant that sizes the slot array.

**Correction: the first version of this PR set that budget to 4/second, and this description claimed
that was the fix. It was wrong, and the follow-up commit `f97a968` raises it to `MAX_GATEWAYS`.** The
limiter is a single *global* window, not per-gateway, so a fleet cold start in which more than four
gateways handshake in the same second gets the surplus refused — and a refusal is terminal, not a
back-off. The server replies `UNAVAILABLE <retry_after>`:

```rust
let unavailable_msg = format!(
    "{session_id} gateway-{gateway_id} UNAVAILABLE {}",
    HANDSHAKE_RATE_WINDOW.as_secs()
);
```

and the client's `perform_handshake` turns that straight into a terminal error and returns:

```rust
CoreResponse::Unavailable { retry_after_seconds } => Err(GatewayError::CoreError(format!(
    "Core unavailable, retry after {retry_after_seconds} seconds"
))),
```

Nothing in `vex-networking`'s client acts on `retry_after_seconds`; `start()` propagates the error and
the gateway never connects. The same ceiling made `cargo xtask test e2e --clients N` unusable for
N > 4 — the surplus clients were refused during startup and never came back.

`MAX_GATEWAYS` is the right number because it is the number the rest of the path is already built
for: `pending_handshakes` is a `[Option<PendingHandshake>; MAX_GATEWAYS]`, so it can already hold a
concurrent in-flight handshake for every gateway id. Every gateway connecting at once is a supported
state, not an attack signature.

## What this does not fix

**A determined attacker can still hold every slot.** 16 allocations/second against 16 slots with a
30 s timeout is far more than enough to keep them all occupied indefinitely — and so was 4, which is
why lowering the budget bought nothing worth the cold-start failures. This bounds resource growth and
lets slots recover once an attacker stops — it does not make the exchange available while one is
active.

The actual fix is **#114: the handshake performs no authentication at all**. Until a credential is
required, anything here is rate-shaping an unauthenticated endpoint. That is deliberately out of
scope: it needs a credential scheme agreed across both repositories.

## Tests

Synthetic-time unit tests for the pure logic — slot reservation rejecting duplicate and
out-of-range ids, only `AwaitingImage` slots expiring at the timeout, and the rolling-window rate
limiter. The Aeron interaction itself is not unit-testable without a live media driver.

The rate-limiter test now fills the window with `MAX_GATEWAYS` allocations rather than with
`HANDSHAKE_RATE_LIMIT` allocations. The original form was self-referential — it asserted the limiter
admits exactly `HANDSHAKE_RATE_LIMIT` requests, which passes for *any* value of that constant,
including 4. Anchoring it to the slot count is what makes it a test of the intended behaviour rather
than of the constant's own definition.

`cargo test -p vex-networking`: 6 passed, 0 failed. The follow-up adds no tests; it changes one
expectation, as described above.

CI clippy is red on this branch, on `networking/src/server/mod.rs:504`
(`useless_borrows_in_formatting`). That file is not touched by this PR and the lint reproduces on
`develop` — it is pre-existing, not a regression from these changes.

## Related

Closes #125

- vex-core #114 — the handshake requires no credential. That is the real fix for squatting; this
  is containment.
- vex-core #123 / PR #155 — the ingress thread rejecting rather than dropping when the ring is full.
  Both are about that thread staying responsive.
- vex-core #130 / PR #165 — container health, which cannot observe a stalled ingress thread.

### DIFF
--- a/networking/src/server/duologue.rs
+++ b/networking/src/server/duologue.rs
@@ -1,4 +1,5 @@
 use crate::server::cmd_handler::FragmentHandler;
+use crate::server::gateway_manager::GatewaySessionEvent;
 use common::OrderCommand;
 use disruptor::{MultiProducer, SingleConsumerBarrier};
 use rusteron_archive::{
@@ -99,6 +100,7 @@ impl Drop for Duologue {
 pub struct DuologueImageAvailable {
     pub expected_session_id: i32,
     pub gateway_id: u8,
+    pub tx: Sender<GatewaySessionEvent>,
 }
 
 impl AeronAvailableImageCallback for DuologueImageAvailable {
@@ -139,6 +141,17 @@ impl AeronAvailableImageCallback for DuologueImageAvailable {
                 session = format_args!("{:#x}", session_id),
                 address = %address
             );
+            if let Err(e) = self.tx.send(GatewaySessionEvent::ImageAvailable {
+                gateway_id: self.gateway_id,
+                session_id,
+            }) {
+                error!(
+                    target: "gateway_manager",
+                    action = "image_available_notification_failed",
+                    gateway_id = self.gateway_id,
+                    error = %e
+                );
+            }
         }
     }
 }
@@ -146,7 +159,7 @@ impl AeronAvailableImageCallback for DuologueImageAvailable {
 pub struct DuologueImageUnavailable {
     pub session_id: i32,
     pub gateway_id: u8,
-    pub tx: Sender<u8>,
+    pub tx: Sender<GatewaySessionEvent>,
 }
 
 impl AeronUnavailableImageCallback for DuologueImageUnavailable {
@@ -162,7 +175,10 @@ impl AeronUnavailableImageCallback for DuologueImageUnavailable {
             session = format_args!("{:#x}", self.session_id)
         );
 
-        if let Err(e) = self.tx.send(self.gateway_id) {
+        if let Err(e) = self.tx.send(GatewaySessionEvent::ImageUnavailable {
+            gateway_id: self.gateway_id,
+            session_id: self.session_id,
+        }) {
             error!(
                 target: "gateway_manager",
                 action = "core_publication_cleanup_request_failed",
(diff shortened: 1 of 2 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

Issue #81, opened 2025-10-22
fix: replay should not have constant frame size


rahulghangas:
We shouldn't be using a constant `FRAMESIZE` for replay, advance by real frame length

PR #175, opened 2026-08-15
fix(replay): recover the journal correctly, or fail loudly instead of booting empty

rahulghangas:
## The problem

Four independent defects, any one of which breaks recovery.

**(a) Live recordings were skipped.** `ActiveRecordingReader` tested `stop_position == 0`, but Aeron
writes `NULL_POSITION = -1` for a live recording — that branch was dead code. `RecorderDescriptorReader`
required `stop_position > 0` and so also skipped a recording left live by a crash, which is exactly
the recovery case. `start_replay` then logged `no_recording_available` at **info** and returned
`Ok(None)`: the exchange booted with an empty book and zero balances, reporting success.

**(b) The replay was bounded by the wrong counter.** `AeronArchiveReplayParams::new`'s first argument
is `bounding_limit_counter_id` and the code passed `0`. Counter id 0 in the C media driver is
`AERON_SYSTEM_COUNTER_BYTES_SENT`, so recovery was bounded by total UDP bytes sent since driver
start.

**(c) A fallback that matched an error string.** The `"recording exists"` path returned a dummy
`subscription_id = 0` with a bare `aeron:ipc` channel and no term positioning, so a new recording was
opened instead of the old one extended — and `shutdown()` then deliberately never stopped it.

**(d) The replay pump dropped commands silently.** `try_publish` on a full ring, with the position
advancing anyway, so recovery could skip journal entries and report success.

## The fix

- `is_live_recording` tests against the binding's `AERON_NULL_POSITION`, and a live recording is
  waited out rather than replayed while live — see below.
- `bounding_limit_counter_id` is `AERON_NULL_COUNTER_ID` — the binding's own unbounded sentinel, not
  a literal.
- The string-matched fallback is deleted; the real error propagates.
- The pump uses blocking `publish`. Replay runs at startup before any traffic is served, so blocking
  for ring capacity is safe there — unlike the live ingress path — and there is a comment saying so.

**A live recording is waited out, not replayed.** The first version of this PR replayed a live
recording up to `get_max_recorded_position` and then handed the descriptor to `start_recording`,
which calls `extend_recording` — and Aeron will not extend a recording that is still active. The
whole journal was replayed only to be rejected at the very last step. The liveness check now runs
**before** `start_replay`: `wait_for_recording_stop` re-reads the descriptor every 100ms until
`stop_position` is no longer `NULL_POSITION`, for at most 10s, then fails with
`live recording {id} was not released within 10 seconds`. The recording is never stopped from here —
a still-live recording may belong to another instance that is running right now, and stopping it
would cut that instance's journal short. Waiting, and then giving up, is the only safe option.
`get_max_recorded_position` is no longer called anywhere.

**An empty journal is not a corrupt one.** The first version of this PR got this wrong as well: it
treated every skipped recording as evidence of corruption, so an instance restarting cleanly after
having recorded nothing (`start_position == stop_position`) refused to boot with
`archive contains 1 recording(s), but none are replayable`. `stop == start` is an empty journal and
is a legitimate state; `stop < start` is corrupt. The two are now counted apart —
`RecorderDescriptorReader` tracks `skipped_empty` and `skipped_invalid`, both logged on
`recordings_listed`, and `ensure_replayable_recording` takes `skipped_invalid` alone and fails only
on that. The selected recording gets the same treatment after the release wait:
`stop_position == start_position` logs `empty_recording_skipped` and returns `Ok(None)`, while
`stop_position < start_position` is an error naming the recording and both positions.

**Recovery otherwise fails loudly instead of booting empty:**

- an archive holding a recording with invalid positions is an error, not an `info` log;
- an empty archive — and an archive whose recordings are all empty — is still a legitimate first
  start and succeeds;
- a decode failure aborts immediately with the recorded message, rather than leaving the loop to spin
  without advancing;
- replay is complete only when the consumed **position reaches the recording's stop position**.

That last point came out of review and matters. The first attempt asserted on a command count derived
as `(stop_position - start_position) / FRAMESIZE`. Aeron inserts padding frames at term boundaries
which consume position but are never delivered to the fragment handler, so on any recording that
crossed a term boundary that count exceeds the number of commands actually replayed — and the check
would have refused to start after a **successful** recovery. A check that fires on correct behaviour
is worse than no check. Position-reached is true regardless of padding. The per-poll check still
compares handler-delivered fragments against published commands, where both sides are genuine counts.

Nine unit tests, still including the padding case — a byte range that is not an exact multiple of
`FRAMESIZE` must still pass. The empty-journal correction **changed an existing test expectation**:
`empty_archive_is_allowed_but_non_empty_archive_without_replay_is_not` asserted
`ensure_replayable_recording(1, false).is_err()`, which is precisely the rule that refused to boot on
an empty journal. That test is gone, replaced by `empty_archive_is_allowed`,
`archive_with_only_empty_recordings_is_allowed`, `corrupt_recording_is_fatal` and
`live_descriptor_requires_release_before_replay`; `ensure_replayable_recording` changed signature
with it.

CI `Build and Test` on `5ac2a43`: `vex-networking` lib 12 passed, `vex-server` lib 7 passed, 0
failed. Two jobs are red and neither comes from this branch — clippy on
`useless_borrows_in_formatting` at `networking/src/server/mod.rs:517`
(`format!("{}?session-id={}", &RECORDING_CHANNEL, replay_session_id)`, untouched here), and the
end-to-end job on `no associated function or constant named new_with_no_credentials_supplier found
for struct AeronArchiveContext`. Both fail identically on `develop` at `2852560`.

## Scope

`FRAMESIZE`-based position advancement is unchanged — that is #81, and this deliberately does not
touch it.

## Related

Closes #131, closes #132

- vex-core #81 — constant frame size and position accounting.
- vex-core #116 / PR #159 — archive write durability. Replay can only be as good as what was written.
- vex-core #127 / PR #158 — journaling ordering, without which the archive content is
  non-deterministic.

### DIFF
--- a/networking/src/server/cmd_handler.rs
+++ b/networking/src/server/cmd_handler.rs
@@ -52,6 +52,8 @@ impl AeronFragmentHandlerCallback for FragmentHandler {
 pub struct ReplayFragmentHandler {
     pub gateway_id: u8,
     pub producer: MultiProducer<OrderCommand, SingleConsumerBarrier>,
+    pub commands_published: i64,
+    pub replay_error: Option<String>,
 }
 
 impl AeronFragmentHandlerCallback for ReplayFragmentHandler {
@@ -60,6 +62,7 @@ impl AeronFragmentHandlerCallback for ReplayFragmentHandler {
             let values = match header.get_values() {
                 Ok(values) => values,
                 Err(e) => {
+                    self.replay_error = Some(format!("failed to decode replay header: {e}"));
                     error!(
                         target: "replay_fragment",
                         gateway_id = self.gateway_id,
@@ -92,18 +95,15 @@ impl AeronFragmentHandlerCallback for ReplayFragmentHandler {
                     "processing replay order command"
                 );
 
-                if let Err(e) = self.producer.try_publish(|cmd| {
+                // Replay runs before live traffic is served, so blocking for ring-buffer capacity
+                // is safe and prevents recovery from silently dropping journal commands.
+                self.producer.publish(|cmd| {
                     *cmd = order_command.clone();
-                }) {
-                    error!(
-                        target: "replay_fragment",
-                        gateway_id = self.gateway_id,
-                        error = %e,
-                        "failed to publish replay order command to ring buffer"
-                    );
-                }
+                });
+                self.commands_published += 1;
             }
             Err(e) => {
+                self.replay_error = Some(format!("failed to decode replay order command: {e:?}"));
                 error!(
                     target: "replay_fragment",
                     gateway_id = self.gateway_id,
--- a/networking/src/server/mod.rs
+++ b/networking/src/server/mod.rs
@@ -39,11 +39,14 @@ use crate::server::gateway_handler::{
 };
 use crate::server::gateway_manager::GatewayManager;
 use crate::server::replay::{
-    ActiveRecordingReader, ExtendedRecordingDescriptor, RecorderDescriptorReader,
+    ExtendedRecordingDescriptor, RecorderDescriptorReader, RecordingCounter,
+    assert_replay_complete, assert_replay_position_complete, ensure_replayable_recording,
+    fail_on_replay_error, is_live_recording,
 };
 use crate::utils::{new_publication_with_mdc, new_subscription_with_handlers};
 use common::{FRAMESIZE, OrderCommand};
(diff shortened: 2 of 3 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #47, opened 2025-08-26
feat: add gateway core networking

esciiee:
## Summary by CodeRabbit

- New Features
  - Aeron-based networking: gateway client and core server with handshake, dedicated channels, message polling, and gateway management.
  - SBE OrderCommand support: schema, generated bindings, encoding/decoding APIs, and runtime crate.
  - Configuration system: environment-aware config loader, validation, merging, logging config, and examples.
  - Docker workflow to generate/extract Rust SBE bindings.

- Tests
  - End-to-end client/server integration test and comprehensive configuration tests/examples.

- Chores
  - CI: added toolchain and cache setup.
  - Workspace: added new workspace members and shared dependencies.

PR #176, opened 2026-08-15
fix(networking): actually authenticate the gateway handshake

rahulghangas:
## The problem

```rust
fn authenticate_gateway(&self, gateway_id: u8, _credentials: &str) -> Result<(), ServerError> {
    // TODO: Implement proper authentication
    info!(target: "gateway_manager", action = "gateway_authenticated", gateway_id);
    Ok(())
}
```

Anyone who could reach the port could claim a gateway slot and submit orders as **any `user_id`**.
The handshake already carried a credential — `HELLO gateway-<id> <key>`, parsed and passed straight
into this function — and it was discarded. The key was used only as an XOR mask on the session id
handed back. The gateway sent a **random** `i32`, so there was nothing to check it against.

## The fix

One configuration field: `gateway_networking.gateway_authentication_key`. It is unavoidable — you
cannot authenticate without a credential, and neither `CoreNetworkingConfig` (which had an
`enable_authentication` flag but no expected value) nor the gateway config carried one. It is a
redacted type, so it cannot be printed by a `Debug` derive.

- The presented credential is compared in **constant time** — fixed-width, all four bytes examined,
  no early exit — so the comparison is not a timing oracle. No dependency was added for it; the
  constant-time helpers in the tree were only reachable transitively.
- **Authentication happens before anything is allocated.** A rejected handshake takes no slot,
  leaves no state, and returns a generic rejection that does not reveal whether the gateway id was
  valid. The rejection is logged at `warn` with the source so an operator can see an attack; the
  credential is never logged.
- The client sends the configured key instead of a random one, and the XOR masking of the session id
  is preserved.
- **It cannot be silently disabled in production.** An empty key, or authentication disabled, is a
  hard startup failure when the environment is production. Development and test may run
  unauthenticated and log loudly that they are.

## Honest limits

This raises the bar from *nothing* to *must know the key*, which is worth having. It is not strong
authentication:

- The credential is a **32-bit integer sent in cleartext** inside the HELLO message over UDP. An
  attacker who can observe the network reads it and replays it. It defends against an attacker who
  cannot see the traffic; it does not defend against one who can.
- 32 bits is also thin against offline guessing if an attacker can attempt handshakes freely — the
  rate limit added in #125 / PR #174 is what bounds that.

The real remedy is an authenticated, encrypted transport — DTLS, or an authenticated key exchange
with AEAD and replay protection. That is a protocol change across both repositories and is
deliberately not attempted here.

## Cross-repo requirement

**`vex-gateway` must send the same key.** Its handshake becomes `HELLO gateway-<id> <configured-key>`
instead of a random `i32`. Deploying this engine change without configuring the gateway locks every
gateway out. The client in `networking/src/client/` is updated here; the gateway's own configuration
is a companion change.

## Scope

Deliberately left alone, despite being mentioned in the same issue:

- **Deposit authorisation.** `DepositFunds` still has no authorisation, no idempotency key and no
  check against configured assets.
- **The `0.0.0.0` bind override** in `src/main.rs` — that is #144, and narrowing it needs to be
  considered against the container deployment.
- **Slot squatting** — #125 / PR #174, which is the rate limit and reaper.

`cargo test -p vex-networking -p vex-config`: 39 passed, 0 failed. clippy: 0 warnings.

## Related

Closes #114

- vex-core #125 / PR #174 — squatting containment; the rate limit is what bounds guessing here.
- vex-core #144 — the unconditional `0.0.0.0` bind.
- vex-gateway #47 — key material handling on the other side of this link.

### DIFF
--- a/.claude/pr68_comments_checklist.md
+++ b/.claude/pr68_comments_checklist.md
@@ -1,0 +1,138 @@
+# PR #68 Unresolved Comments Checklist
+
+Generated: 2026-01-20
+Updated: 2026-01-20 (Session 2)
+
+## Summary
+- **Total unresolved comments:** 59
+- **Source:** All from `coderabbitai` bot
+- **Status:** All Critical (P1) items addressed, most P2/P3 items addressed
+
+## Priority Legend
+- **Critical (P1):** Must fix - potential bugs, compilation errors, security issues
+- **Major (P2):** Should fix - code quality, reliability issues
+- **Minor (P3):** Nice to have - typos, documentation, style
+- **WNF:** Will Not Fix - not relevant, already fixed, or disagree
+
+---
+
+## Critical Priority (P1) - ALL ADDRESSED
+
+- [x] **common/src/order.rs:2512805059** - Invert PriceCache sentinels to match book semantics ✅ FIXED
+- [x] **common/src/order.rs:2512805080** - Fix `PriceCache::new` signature ✅ WNF (code compiles fine)
+- [x] **common/src/l2_market_data.rs:2512948095** - Initialize level buffers ✅ WNF (false positive)
+- [x] **vex-config/src/networking.rs:2512948127** - Clamp gateway IDs to configured capacity ✅ FIXED
+- [x] **xtask/Cargo.toml:2512948131** - Verify if test_e2e binary should remain commented out ✅ WNF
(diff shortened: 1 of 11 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #177, opened 2026-08-15
perf: remove quadratic event appends, per-order allocations and hot-path logging

rahulghangas:
Four independent hot-path findings.

## 1. `attach_event` was O(n²) per order

It walked the event chain from the head for every appended fill. A 500-order sweep took
**235 µs** against a documented 50–70 µs p99 target.

`OrderCommand::events` becomes a `MatcherTradeEventChain` that keeps a tail pointer, so appending is
O(1) amortised and the whole sweep is O(n).

**Measured, release mode, same 500-order sweep shape, 100 samples:**

| | Before | After |
|---|---:|---:|
| Sweep median | 235.2 µs | **25.4 µs** |
| Append path alone | 196.5 µs | **14.8 µs** (13.3×) |

The `Box` per fill stays — `next_event: Option<Box<_>>` is the public ownership model and changing it
is a wider change than this warrants.

**One correction came out of review and is worth recording.** The first version took the tail as the
address of the incoming box, assuming it was a single node. `MatcherTradeEvent::next_event` is a
`pub` field and code constructs multi-node events directly (`processors/src/events.rs:794`), so
appending one of those would have left `tail` pointing at its head — and the next append would
overwrite that link, **silently dropping every trade after it** from the chain that feeds the `trades`
topic. `new` and `push` now locate the real tail of whatever they are given. `find_tail` is iterative,
not recursive, so it does not add a second stack-overflow path alongside the recursive `Drop` already
noted in #141.

## 2. `record_snapshot` allocated per order on the matching thread

Now only the populated price and volume vectors are allocated; `ask_orders` and `bid_orders` get zero
capacity. The `SystemTime` `unwrap()` is gone — a clock failure yields timestamp `0` rather than
killing the matching thread.

Codex's investigation corrected the issue text here: **two** of the vectors are always empty, not
four. Recorded so the count is not repeated as fact.

## 3. INFO logging per command on pinned pipeline threads

`cmd_handler.rs:17` `Debug`-formatted the entire `OrderCommand` for every inbound fragment;
`events.rs:531` and `journaling.rs:40` logged per command. All three serialise on the global stdout
mutex, on core-pinned threads.

All three demoted to `debug!`, so they remain available but are off by default. **Logs that report an
actual failure keep their level** — nothing that reports a problem was quietened.

## 4. `PriceLevel::total_volume` used unchecked `+=` / `-=`

It wrapped in release and panicked in debug. All three aggregate updates are now explicit:
`PriceLevel::add_order` (`orderbook/src/lib.rs:88`), `PriceLevel::remove_order`
(`orderbook/src/lib.rs:104`), and the maker-fill decrement inside `match_order`
(`orderbook/src/lib.rs:211`).

**The first version of this PR chose the wrong failure mode, and an earlier revision of this
description defended it as deliberate.** That version used `checked_add` / `checked_sub` with
`.expect("price level total volume overflow" / "underflow")`, and argued that panicking in release was
correct because a volume mismatch proves the book's accounting has already diverged from its orders.
Codex's review on `orderbook/src/lib.rs:69` showed that argument does not hold for the overflow side:

> When two balance-backed resting orders at the same price have sizes whose sum exceeds `u64::MAX`,
> both orders can pass the per-user risk checks, but the second order now reaches this `expect` and
> panics during order-book processing. Aggregate overflow does not imply that the book has already
> diverged, so it should be handled by a wider accumulator or by rejecting the incoming order rather
> than terminating the processing thread.

That is right. Two individually risk-approved orders resting at one price can sum past `u64::MAX`
without anything being corrupt — it is an aggregate-only limit, and killing the matching thread for it
converts a bookkeeping ceiling into an outage.

**Rejecting the incoming order, the reviewer's other suggestion, is not reachable from this call
site**, and the reason matters more than the mechanics:

- `add_to_book` takes `cmd: &OrderCommand` immutably (`orderbook/src/lib.rs:403`), so it cannot set a
  status at all.
- Even if it could, `place_order` overwrites the status on the very next line —
  `cmd.set_status(Status::Placed)` / `Status::PartiallyFilled`.
- Decisively: `Status::Rejected` does not release reserved funds. `create_risk_r2_handler!` returns
  early on `Rejected` before it reaches `handle_cancellation` (`server/src/utils.rs:64`). A reject
  raised here would leave the user's collateral reserved with no order behind it. Silently stranding
  money is a worse outcome than the panic it was meant to replace.

So all three sites saturate — `saturating_add` at `u64::MAX`, `saturating_sub` at zero — and emit a
structured error line naming the price level, the existing total and the operand, e.g.

```
level=error event=price_level_volume_overflow price_level=100 existing_total=18446744073709551615 addend=1 message="price level total volume overflowed; saturated at u64::MAX"
```

`vex-orderbook` depends only on `common` and pulls in no logging facade, so this is a small local
`error!` macro over `eprintln!` rather than a new dependency. It fires only on the failure path, so it
adds nothing to the hot path. The condition stays loud and greppable without taking the exchange down
for it; **this PR no longer panics in release on volume over/underflow.**

The `#[should_panic]` test the first commit added (`price_level_volume_underflow_is_loud`) went with
the panic. It is replaced by `price_level_volume_overflow_saturates_without_panicking` and
`price_level_volume_underflow_saturates_without_panicking`, which assert the saturated totals, the
resulting order counts, and that `remove_order` still reports `Status::Cancelled`.

## Verification

`cargo test --workspace`: 162 passed, 0 failed. clippy: 0 warnings. Against `develop`, **no existing
test body was changed** — `orderbook/src/unit_tests.rs` is additions only, and the one test that was
rewritten was added by this PR's own first commit. The observable matching behaviour is identical.

## Related

Closes #142

- vex-core #141 item 2 — unbounded recursive `Drop` on the event chain, deliberately untouched.
- vex-core #79 — snapshot allocation, referenced by item 2.
- vex-core #51 — logging levels, referenced by item 3.
- vex-core #129 / PR #156 — the panic containment that item 4's first version leaned on to justify
  aborting. Item 4 no longer panics, so it no longer depends on it.

### DIFF
--- a/common/src/cmd.rs
+++ b/common/src/cmd.rs
@@ -9,6 +9,8 @@ use sbe_order::{ReadBuf, SBE_SCHEMA_ID, SbeResult, WriteBuf};
 use serde::de::Error;
 use serde::de::value::Error as SerdeError;
 use serde::{Deserialize, Serialize};
+use std::fmt;
+use std::ops::{Deref, DerefMut};
 
 // Size of the serialized OrderCommand in bytes
 // Header: 8 bytes
@@ -81,7 +83,7 @@ pub struct OrderCommand {
     pub status: Status,
 
     /// Passive order linked list
-    pub events: Option<Box<MatcherTradeEvent>>,
+    pub events: Option<MatcherTradeEventChain>,
 
     /// Final balance of the user/maker after this trade
     /// For Deposit/Withdraw commands, only balance[0] is used to represent the asset balance change
@@ -248,7 +250,7 @@ impl OrderCommand {
     }
 
     pub fn events_mut(&mut self) -> Option<&mut MatcherTradeEvent> {
-        self.events.as_mut().map(|e| e.as_mut())
+        self.events.as_deref_mut()
     }
 
     pub fn set_size(&mut self, size: u64) {
@@ -268,13 +270,10 @@ impl OrderCommand {
     }
 
     pub fn attach_event(&mut self, event: Box<MatcherTradeEvent>) {
-        if let Some(mut tail) = self.events.as_mut() {
-            while tail.next_event.is_some() {
-                tail = tail.next_event.as_mut().unwrap();
-            }
-            tail.next_event = Some(event);
+        if let Some(events) = self.events.as_mut() {
+            events.push(event);
         } else {
-            self.events = Some(event);
+            self.events = Some(MatcherTradeEventChain::new(event));
         }
     }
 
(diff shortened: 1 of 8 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #178, opened 2026-08-15
fix(ledger): consistent price snapshots, zero-price rejection, honest cancel status, bounded balance map

rahulghangas:
Six grouped ledger and settlement findings. Four fixed, one skipped as already fixed elsewhere, one
declined with reasons.

## 1. Truncation residue permanently locked — **skipped, already fixed**

Fixed by the open PR for #169, which tracks the exact amount locked per BID order and releases that
instead of recomputing it. Deliberately not reimplemented here — a second mechanism for the same
invariant is worse than one.

## 2. `PriceCache` torn reads and a panic on a missing market — **fixed**

`update_prices` wrote `best_bid` and `best_ask` as two independent atomics, so a reader could observe
one market's bid paired with another update's ask — a price pair that never existed. It also
`unwrap()`ed on a missing market.

`MarketPrice` gains a **sequence counter** (a seqlock): the writer makes it odd for the duration of
the update, and `get_prices` re-reads it after loading both values and retries if it changed. A
reader therefore only ever sees a pair from one completed update. **No lock was added** — the
matching path stays lock-free, which was the constraint. A missing market is now a no-op rather than
a panic.

`get_best_bid` and `get_best_ask` are unchanged in behaviour; they now read through the consistent
snapshot.

## 3. R1 locks for order n+1 before R2 settles order n — **declined, with reasons**

The reinvestment guarantee this asks for requires `R2(n) ≺ R1(n+1)`. That collapses the pipeline's
overlap: R1 for every subsequent order would wait on R2 of the previous one, serialising the stages
the disruptor exists to run concurrently. There is no local ledger fix — the ordering *is* the
design.

The same reasoning applies to the scheduling half of item 2: making a market BUY's lock price
deterministic across a replay would need `Matching(n) ≺ R1(n+1)` or journaling moved, which is the
same restructuring.

Both are genuine, and both are pipeline-topology decisions rather than bugs to patch. Recorded here
so the trade-off is on the record rather than rediscovered.

## 4. `bid_amount` guarded only the `u64::MAX` sentinel — **fixed**

A `best_ask` of 0 made every market BUY lock zero collateral. Both sentinel values are now rejected
through the existing `InvalidArguments` path — no new error type.

## 5. `handle_cancellation` ignored a failed release — **fixed**

The client was told `Cancelled` while the funds stayed locked. A failed release now sets
`Status::Rejected`, so the client is not told its order was cancelled when it was not.

## 6. `BalanceStore::get_balance_mut` inserted before validation — **fixed**

`entry().or_default()` created a balance row for any user id that reached it, so rejected orders for
non-existent users grew the map without bound — an unauthenticated memory-growth vector while #114
was open. A missing balance now returns `UserAssetNotFound` and inserts nothing. `lock_funds` and
`unlock_funds` use checked arithmetic and leave the balance untouched on overflow rather than
wrapping.

## Verification

`cargo test -p processors -p vex-server -p common`: 46 passed, 0 failed. clippy: 0 warnings.
Includes a torn-read test that fails against two independent atomics.

## Related

Closes #140

- vex-core #169 / PR #171 — item 1.
- vex-core #114 / PR #176 — item 6's growth vector is only reachable pre-authentication.
- vex-core #128 — the pipeline aliasing question, which item 3's ordering discussion touches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when reading bid and ask prices during concurrent updates.
  * Missing market symbols are now handled safely without errors.
  * Balance operations report missing assets correctly and prevent changes when an overflow would occur.
  * Failed cancellations are marked as rejected.
  * Market-buy reservations now reject zero or unavailable reference prices.
* **Improvements**
  * Added explicit support for setting or replacing account balances.
  * New balance records are created only during valid fund additions.

### DIFF
--- a/common/src/order.rs
+++ b/common/src/order.rs
@@ -44,6 +44,7 @@ impl Order {
 
 // Holds the top-of-book prices for a single symbol
 pub struct MarketPrice {
+    sequence: AtomicU64,
     pub best_bid: AtomicU64,
     pub best_ask: AtomicU64,
 }
@@ -54,6 +55,7 @@ impl Default for MarketPrice {
         // - No bid means price floor (0) - can't sell
         // - No ask means price ceiling (MAX) - can't buy
         Self {
+            sequence: AtomicU64::new(0),
             best_bid: AtomicU64::new(0),
             best_ask: AtomicU64::new(u64::MAX),
         }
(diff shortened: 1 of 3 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #179, opened 2026-08-15
fix(networking): bound gateway ids, drop event chains iteratively, timeout startup, size ports correctly

rahulghangas:
Six grouped Aeron transport findings. Four fixed, one skipped as already fixed, one declined.

## 1. `Publications` indexed a fixed array by `gateway_id` with no bound check — **fixed**

Every current caller happened to be guarded, but the invariant was unenforced at the API boundary —
one unguarded caller away from an index panic on the ingress thread. `set`, `get` and `remove` now
reject `gateway_id >= MAX_GATEWAYS` and return an error rather than indexing.

## 2. Unbounded recursive `Drop` on the event chain — **fixed**

When a ring slot was reused, the previous lap's `MatcherTradeEvent` chain was freed by recursive
`Drop` **on the Aeron ingress thread**. A deep chain — a large sweep across many price levels —
could overflow the stack and take the process down at exactly the moment it was busiest.

`Drop` is now iterative: it walks the chain taking each `next_event` in a loop. A regression test
drops a 100,000-node chain, which the recursive version cannot survive.

## 3. Unbounded startup wait and an error storm in the poll loop — **fixed**

The archive and replay startup waited on `is_connected()` with **no timeout**, so a misconfigured
archive hung the engine at boot with no diagnosis. It now times out after a private 10-second
constant and fails with a clear error.

The main poll loop re-logged errors at busy-spin frequency — millions of lines a second for a
persistent fault, which fills a disk and hides everything else. The first occurrence is logged and
subsequent identical errors are suppressed; a recovery resets the de-duplication so a **new**
occurrence is reported. Nothing is lost, the storm is.

## 4. Port allocator sized for one port per gateway — **fixed**

`PortAllocator` was sized with `max_gateways`, but each gateway consumes **two** ports (data and
control), so effective capacity was half the configured maximum — the engine would run out of ports
at 8 gateways with `MAX_GATEWAYS = 16`. Capacity is now `2 × max_gateways`.

`SessionAllocator::new` also computed `max - min` unchecked and overflowed on an inverted range. It
now rejects inverted or overflowing ranges instead.

## 5. `send_message_with_retries` sleeping in the poll callback — **skipped, already fixed**

The only server-side caller is the handshake path, which the open PR for #125 made non-blocking. The
identically-named client method is a different call site and is not on the server poll callback.
Verified rather than assumed; no second fix written.

## 6. The dedicated per-gateway channel is guarded only by a session id — **declined, with reasons**

The session id is handed back in a plaintext handshake response, so anyone who observes it can use
the channel. Closing this needs an authenticated, encrypted transport — key handling, integrity and
replay protection, and a compatible protocol change on both sides.

The shared key added by #114 / PR #176 does not cover it: that authenticates the *handshake*, and
the channel itself stays plaintext. Worth being explicit, because it would be easy to assume #114
closed this.

## Verification

`cargo test -p vex-networking -p common`: 25 passed, 0 failed. clippy: 0 warnings. The SBE wire
format and `ORDERCOMMANDSIZE = 64` are unchanged.

## Related

Closes #141

- vex-core #125 / PR #174 — item 5, and the same ingress-thread-must-not-block principle.
- vex-core #114 / PR #176 — handshake authentication, which does **not** close item 6.
- vex-core #142 / PR #177 — also touches `common/src/cmd.rs`, changing `OrderCommand::events` to a
  chain type. The two are semantically compatible — the iterative `Drop` here applies to the same
  nodes — but expect a textual conflict.

### DIFF
--- a/common/src/cmd.rs
+++ b/common/src/cmd.rs
@@ -355,6 +355,15 @@ pub struct MatcherTradeEvent {
     pub maker_original_size: u64,
 }
 
+impl Drop for MatcherTradeEvent {
+    fn drop(&mut self) {
+        let mut next = self.next_event.take();
+        while let Some(mut event) = next {
+            next = event.next_event.take();
+        }
+    }
+}
+
 impl MatcherTradeEvent {
     pub fn calc_filled_size(&self) -> u64 {
         let mut size = 0;
@@ -409,4 +418,21 @@ pub fn decode_order_command(buf: &[u8]) -> Result<OrderCommand, SerdeError> {
         route_gateway_id: 0,
         original_size: 0,
     })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deep_matcher_trade_event_chain_drops_iteratively() {
+        let mut chain = None;
+        for _ in 0..100_000 {
+            let mut event = MatcherTradeEvent::default();
+            event.next_event = chain;
+            chain = Some(Box::new(event));
+        }
+
+        drop(chain);
+    }
 }
--- a/networking/src/server/gateway_manager.rs
+++ b/networking/src/server/gateway_manager.rs
@@ -33,6 +33,12 @@ enum GatewaySlotState {
     Live,
 }
 
+const PORTS_PER_GATEWAY: usize = 2;
+
+fn gateway_port_capacity(max_gateways: u16) -> usize {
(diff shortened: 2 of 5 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #180, opened 2026-08-15
fix(config): gate balance preload, honour the configured broker, surface disabled journaling

rahulghangas:
Three grouped balance-preload and configuration findings.

## 1. `balance-preload` had no environment gate — **fixed**

The feature deposits balances at startup, **before the archive publication exists**, so preloaded
balances were never journaled and vanished on replay — the exchange would come back with different
balances than it shut down with. And a cargo feature is a **build-time** choice, not a deployment
guard: a binary built with it preloads in production.

`validate()` now rejects `balance_preload` in a production environment, and the server validates
before initialisation so the deposits never execute there.

Fail-fast rather than making the deposits journaled: preloading balances is a development and test
affordance, and the right answer in production is that it does not happen at all, not that it happens
durably.

## 2. Configuration overwritten in `main.rs` — **partially fixed, deliberately**

`kafka_broker` was discarded in favour of a `KAFKA_BROKER` environment variable that nothing else in
the repository sets, so the configured value could never take effect. The configured value is now
honoured and the ad-hoc override is removed.

**Two overwrites are kept on purpose**, and this is the part worth reviewing:

- `context_dir = "/dev/shm/aeron"` — the shipped compose deployment shares exactly that path with the
  media-driver container. Honouring a different configured value would break it. (See #130 / PR #165,
  where this path alignment is what makes the health check work at all.)
- `local_address = "0.0.0.0"` — a container has to bind an address reachable from outside itself.
  Narrowing it needs to be designed against the deployment, and doing it here would ship a gateway
  that cannot connect.

Both are real findings; neither is safe to fix without also changing the deployment, so they stay
until that is done together.

## 3. An empty `request_control_channel` silently disabled journaling — **fixed**

The exchange could run with **no journal at all** and look completely healthy. `validate()` never
checked it.

Journaling being optional is a supported mode — `networking/src/server/mod.rs:133` logs
`archive_recording = false` as a normal startup, and #116 / PR #159 deliberately relies on that
staying true. So the fix is not to make it fatal everywhere; it is to make it **deliberate and
visible**:

- `validate()` rejects an empty `request_control_channel` in production.
- A non-production startup emits an **ERROR-level** `journaling_disabled` event saying replay is
  impossible.

An operator should never discover there was no journal during a post-incident investigation.

`cargo test -p vex-server -p vex-config`: 43 passed, 0 failed, plus 2 feature-specific
balance-preload tests. clippy: 0 warnings, including with the feature enabled.

## Related

Closes #144

- vex-core #134 / PR #166 — the sibling defect: a config load failure silently becoming zero-fee
  Test defaults.
- vex-core #116 / PR #159 — archive durability, which depends on journaling-optional remaining a
  supported mode.
- vex-core #130 / PR #165 — the compose deployment that item 2's kept overwrites serve.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -73,6 +73,12 @@ impl RunningEngine {
 /// # }
 /// ```
 pub fn start(config: VexConfig, replay: bool) -> Result<RunningEngine, EngineError> {
+    config
+        .validate()
+        .map_err(|error| EngineError::Configuration(error.to_string()))?;
+
+    report_disabled_journaling(&config);
+
     #[allow(unused_mut)] // mut needed when balance-preload feature is enabled
     let ((engine, mut producer), replay_control) = init_internal(
         config.symbols.symbols.clone(),
@@ -117,6 +123,18 @@ pub fn start(config: VexConfig, replay: bool) -> Result<RunningEngine, EngineErr
         thread: thread_handle,
         shutdown_flag,
     })
+}
+
+fn report_disabled_journaling(config: &VexConfig) {
+    if config.core_networking.request_control_channel.is_empty() {
+        tracing::error!(
+            target: "server",
+            action = "journaling_disabled",
+            archive_recording = false,
+            environment = %config.environment,
+            "JOURNALING IS DISABLED: commands will not be recoverable by replay"
+        );
+    }
 }
 
 /// Internal initialization function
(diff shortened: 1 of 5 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #181, opened 2026-08-15
test: one pipeline construction, real fee assertions, honest invariant checks, both CI feature sets

rahulghangas:
Five grouped test-suite and build findings. Four fixed, one declined with reasons.

## 1. The disruptor pipeline was transcribed twice — **fixed**

`build_disruptor_pipeline` and `build_test_pipeline` were two hand-maintained copies of the same
14-stage construction, with nothing keeping them in sync.

There is now **one** construction. The test path joins it by attaching an observer to the final
handler, and pinning is honoured: present means enabled, absent means disabled. The two copies can no
longer drift, because there is only one.

The first cut of this unification got the default wrong, and the correction is worth stating plainly
because this description originally asserted the opposite. Every test call site passed
`false, // Tests don't use pinning` as `enable_pinning`, which was read here as the test-pinning API
having rotted into dead code that "could never have caught a pinning regression". It was dead — but
that hard-coded `false` was also the only thing keeping tests off specific cores. Collapsing the call
sites into `Some(pinning) => (pinning.into(), true)` while `TestEngineBuilder::new()` still defaulted
`core_pinning` to `Some(TestCorePinning { journaling: 1, risk_engines: [2, 3, 4, 5],
matching_engines: [6, 7, 8, 9], risk_r2_engines: [10, 11, 12, 13], events: 14 })` turned an explicit
opt-out into an on-by-default. Every test then tried to pin to cores 1-14, and on any host that does
not have them the disruptor's `cpu_has_core_else_panic` (`affinity.rs:9`) aborts the run:

```
No core with ID=<n> is available.
```

for the first of cores 1-14 the host lacks.

CI never saw this, which is why it needs calling out rather than quietly fixing. All three test jobs
in `ci.yml` export `VEX_ENV: dev`, and `setup_tuple` carried a `VEX_ENV` branch that called
`without_cpu_pinning()` for `dev`/`development`. `Build and Test` passed on the unification commit
(`vex-server` lib: `7 passed; 0 failed`, under both feature sets) purely because of that env var. The
regression was reachable only from a plain `cargo test` on a machine with fewer than 15 cores.

The default is now `core_pinning: None`, and `without_cpu_pinning()` is replaced by an explicit
`with_cpu_pinning(TestCorePinning)` opt-in — pinning is something a test asks for, not something it
has to remember to decline. The `VEX_ENV` branch in `setup_tuple` goes with it: off by default needs
no escape hatch, and an env var that silently changes thread affinity is worse than an argument.
`test_engine_builder_does_not_pin_by_default` asserts the default so it cannot flip back unnoticed.

## 2. `--all-features` shrank the ring to 256 — **fixed**

`test-config` sets the buffer to 256, and CI ran **only** `--all-features`, so the production buffer
size of 1024 and the pipeline built around it were never executed by any test. A back-pressure or
capacity defect at the real size would not have been caught.

CI now runs both: default features, then `--all-features`. The production configuration is tested.

## 3. Tautological fee assertions — **fixed**

The engine tests computed their expected fee with the **production formula**, so they would have
passed whatever that formula was. Several also evaluated to zero because the sizes were too small,
and the shipped configs use zero fees — so the assertions were both circular and vacuous.

Expected values are now **hand-calculated literals**, at sizes where the fee is non-zero, and a
representative test runs at the shipped `100000/10` scale factors. The tests can now fail.

## 4. An unreachable invariant check with documentation that promised more — **fixed**

`verify_no_negative_balances` compared `available > total`, where `total()` is
`available.saturating_add(locked)` — the condition is **unreachable**, so its "wraparound detection"
detected nothing. Meanwhile the doc comment promised balance conservation and trade-balance coupling,
neither of which existed.

The check now detects a genuine `available + locked` overflow via `checked_add`, and the
documentation says what the verifier actually checks. **The conservation claim is removed rather than
implemented** — conservation needs the fee accumulator added in #118 / PR #173, which is not visible
from this branch. Better an honest short list than a doc comment describing a verifier that does not
exist.

## 5. Commented-out end-to-end tests — **declined, with reasons**

`server/tests/integration_test.rs` is not `#[ignore]`d active code — it is **entirely commented out
and stale**. Un-commenting it does not produce a skipped test; it produces a compile failure, because
it is written against APIs that have since changed. Making it run needs a rewrite against current
APIs plus deterministic shutdown and Aeron media-driver lifecycle management.

That is real work with a real design question attached, not a checkbox, so it is left as it is rather
than half-restored.

## Verification

`cargo test --workspace`: 155 passed, 0 failed. `cargo test --workspace --all-features`: 155 passed,
0 failed. clippy: 0 warnings. Those counts are from the unification commit; the pinning follow-up
adds one test (`test_engine_builder_does_not_pin_by_default`) to each.

Note that a green run is weaker evidence here than it looks, per item 1: run the suite without
`VEX_ENV=dev` set, on a host with fewer than 15 cores, if you want the pinning default exercised.

## Related

Closes #143

- vex-core #118 / PR #173 — the fee accumulator that item 4's conservation check would need.
- vex-core #135 / PR #170 — shipped-scale money tests, the same class of gap as item 3.
- vex-core #136 / PR #167 — core-pinning validation, which item 1's dead test-pinning API should have
  been exercising.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,9 @@ jobs:
           RUSTFLAGS: "-D warnings -C link-arg=-fuse-ld=mold"
 
       - name: Run tests
-        run: cargo test --workspace --all-features --all-targets
+        run: |
+          cargo test --workspace --all-targets
+          cargo test --workspace --all-features --all-targets
         env:
           RUSTFLAGS: "-D warnings -C link-arg=-fuse-ld=mold"
           VEX_KAFKA_BROKER: localhost:9092
--- a/processors/src/risk_engine.rs
+++ b/processors/src/risk_engine.rs
@@ -353,7 +353,7 @@ impl RiskEngine {
         // PR #178 made get_balance_mut fallible so a REJECTED operation can no longer insert
         // an empty row. This is the commit point: every validation above has passed, so the
         // write must succeed even for an asset the user is receiving for the first time.
-        // set_balance inserts-or-updates, keeping PR #153's all-or-nothing commit intact.
+        // set_balance inserts-or-updates, which keeps #153's all-or-nothing commit intact.
         if asset_to_add == asset_to_subtract {
             store.set_balance(user_id, asset_to_add, balance_add);
         } else {
@@ -1182,7 +1182,7 @@ mod tests {
         // collateral record is deliberately NOT an error (see
         // test_bid_cancel_without_collateral_record_is_safe). The ASK path still unlocks base
         // directly, so it is the case that genuinely fails and must surface as Rejected --
-        // which is the guarantee PR #178 is about.
+        // which is what PR #178's guarantee is about.
         let mut cmd =
             OrderCommand::place_order(TimeInForce::Gtc, user_id, 100, 10, Side::Ask, market_id, 1);
         cmd.set_status(Status::Cancelled);
--- a/server/src/engine.rs
+++ b/server/src/engine.rs
@@ -654,7 +654,6 @@ pub mod test {
         matching_engines: [usize; NUM_MATCHING_ENGINES],
         risk_r2_engines: [usize; NUM_RISK_ENGINES],
         events: usize,
-        test_handler: usize,
     }
 
     impl Default for TestCorePinning {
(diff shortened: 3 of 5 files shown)


rahulghangas:
@codex review

rahulghangas:
@codex full review

PR #182, opened 2026-08-15
chore: delete dead code, correct misleading docs, and add the missing codec round-trip test

rahulghangas:
Twelve grouped dead-code and doc-drift findings. Nine fixed, one partly, two declined.

## Fixed

**1. `CoreArithmetic` encoded a different money formula.** Dead code computing
`size * (price * quote_scale_k)` — a **multiplication where the live `calculate_quote_cost`
divides**. Two contradictory money formulas in one codebase, one of which nothing calls. Verified the
live divide-based normalisation is the one consistent with the shipped native scales, and deleted the
dead one. **No live formula changed.**

**2. `base_scale_k` documented as the lot-size multiplier, never applied.** Confirmed by grep: it is
stored, validated as non-zero, and read by no calculation anywhere. It is now documented as a
reserved compatibility field, and the shipped and example values move from `100_000` to a neutral `1`
so the configuration stops claiming an effect it does not have. **Behaviour-neutral today** — worth
noting explicitly that if anyone later wires `base_scale_k` up, the ETH/USDT intent of `100_000` will
need restoring deliberately rather than inherited by accident.

**4. Dead and unreachable code removed:** `create_matching_handler!` (which could not compile),
`create_event_handler!`, `journal_event`, `BookSide::iter_for_l2`, `MatcherEventType`,
`common/src/events.rs` in its entirety, the unused `L2MarketData` fields, and the unreachable
`Status::Processed` arm.

**5. Misleading archive log labels.** `publish_to_archive` named a local `gateway_id` that actually
held `cmd.order_id`, so its error logs collided with real gateway ids — an operator grepping for a
gateway would have found order ids. Now labelled `order-{}`.

**6. `PriceCache` doc comments contradicted the code in three places** — `get_best_bid` documented
`u64::MAX` for "unavailable" while the empty-book sentinel is `0`. All three corrected.

**7. The orderbook doc claimed cancellation was a linear scan.** It binary-searches, on an ordering
invariant that was real but **unasserted**. Doc corrected, and a `debug_assert!` added so a future
change to id generation fails loudly instead of silently corrupting cancels. Debug-only — no runtime
cost on the matching path.

**8. The unreachable FOK double-check removed.** `can_fill_completely` already filters same-user
orders, so self-trade prevention cannot leave a remainder after it returns true; the post-match
`remaining == 0` branch could never be false.

**9. `OrderCommand::user_id` docs** still described the obsolete gateway-id-in-`user_id` cancel
contract, superseded by `route_gateway_id`. Corrected.

**12. R1 cloned the shared `Arc<PriceCache>` before the shard check**, contending one refcount across
four pinned threads for **every** order — three of four clones were discarded immediately. The shard
check now comes first, so only the owning shard pays.

## Partly fixed

**10. SBE `blockLength` is 64 but only 56 body bytes are encoded**, and nothing regenerates the codec
from `ordercmd.xml`. The **round-trip test is added** — every field, boundary values,
`decode(encode(x)) == x` — which the codec previously had **none of at all**. The 8 wasted bytes per
message are left: changing `blockLength` is a wire-format change across both repositories.

## Declined

**11. `route_gateway_id` is not on the wire**, so replayed commands all carry gateway 0 into the
snowflake node field. Same reason as above — putting it on the wire is a cross-repo breaking change,
not a cleanup.

**3. `market_type` is never read**, and the development and test BTC/USDT market is declared
`FuturesContract` on a spot-only engine. Reported rather than changed: silently altering a shipped
market's declared type is not a cleanup decision.

## On the FOK path

An earlier draft of this PR rewrote FOK matching into a plan-then-apply structure. It was reverted:
it allocated a `Vec` per order and per price level on the matching hot path — precisely what #142 /
PR #177 removes — and it silently changed the market-order capacity predicate from
`is_market_order(cmd) || price <= cmd.price` to `price <= cmd.price`. That may well be more correct,
but a matching-behaviour change does not belong in a cleanup, and fill-for-fill equivalence was
unproven. Only the unreachable branch is removed here, and the module doc now says plainly that FOK
**still makes two traversals** and that removing the second belongs with #142.

`cargo test --workspace`: 154 passed, 0 failed. clippy: 0 warnings. `ORDERCOMMANDSIZE = 64`,
`FRAMESIZE = 96` and SBE `blockLength = 64` are unchanged.

## Related

Closes #145

- vex-core #142 / PR #177 — the hot-path performance work item 12 belongs beside, and where the FOK
  second traversal should be addressed.
- vex-core #94 — the order-id ordering invariant item 7 now asserts.


rahulghangas:
@codex review

rahulghangas:
@codex full review

Issue #183, opened 2026-08-17
fix: R2 settlement holds aliasing &mut OrderCommand when taker and maker are on different shards


rahulghangas:
Split out of #128, whose other handler groups are fixed by PR (see below). This is the one case that could not be resolved without changing `processors/`.

## Summary

Every other disruptor handler group can establish a single owner before creating a mutable reference:

- **R1** and **matching** read their shard key through a shared reference, return early if they are not the owner, and only then create the `&mut`.
- **Journaling** and **events** are the sole handler in their barrier group.

**R2 cannot.** `create_risk_r2_handler!` (`server/src/utils.rs`) is entered by all four shards, and a shard does work if it owns *either* the taker or any maker in the trade-event chain:

```rust
let is_taker_shard  = ((taker_id & shard_mask) as usize) == $shard_id;
let is_maker_shard  = ((maker_id & shard_mask) as usize) == $shard_id;
```

Taker and maker routinely live on different shards, so two R2 handlers legitimately settle the same command **concurrently**, and both currently do it through `let cmd = unsafe { &mut *cell.get() };`. Two live `&mut` to one object is UB regardless of whether the writes are disjoint.

The code now says so honestly rather than asserting a safety it does not have:

```rust
// SAFETY: R2 is the one handler group where taker and maker may be settled by
// different shards concurrently, so two handlers can hold `&mut` to the same
// command. This access is not sound. Making it sound requires
// `handle_trade_event` to take projected fields rather than `&mut OrderCommand`,
// which is a change in `processors/` and is tracked in issue #128.
```

## Why the obvious fix was rejected

An attempt to project fields — settle into a local `MatcherTradeEvent` and copy `maker_balance` back — was written and **reverted during review**, because it changes settlement behaviour. `settle_trade` (`processors/src/risk_engine.rs:271-279`) writes `event.maker_balance` for **both** participants:

```rust
if is_maker { event.maker_balance[0] = balance_sub; event.maker_balance[1] = balance_add; }
else        { event.maker_balance[0] = balance_add; event.maker_balance[1] = balance_sub; }
```

With only the maker shard copying back, the surviving value depends on whether taker and maker land on the same shard — same shard, the taker's write is last and wins; different shards, the maker's wins. That is new nondeterminism on the settlement path, introduced by what was meant to be a pure soundness fix. It also required punning through `Option<Box<MatcherTradeEvent>>`'s null-pointer layout.

[...]

PR #184, opened 2026-08-17
fix: stop event handlers minting aliasing &mut to the same ring slot

rahulghangas:
> Stacked on #158 (`fix/127-journaling-barrier`) and on **trade-vex/disruptor-rs#3**. Merge the
> disruptor PR and bump this pin to the resulting `main` SHA before merging here — the pin currently
> points at that PR's branch head so this builds and tests today.

## The problem

Every event handler in a barrier group received its own `&mut OrderCommand` to the **same** ring
slot. `spawn_processor` runs once per processor, so four R1 shards, four matching shards and four R2
shards each minted an exclusive reference to one object, concurrently, on different threads.

Two live `&mut` to one object is undefined behaviour by the language rules. `&mut T` carries a
`noalias` guarantee LLVM is entitled to exploit — hoisting loads across writes it believes
unobservable, caching values in registers across another thread's write, reordering. **This holds
whether or not the handlers write disjoint fields**, so "each shard only touches its own users" was
never a defence. It works today; nothing guarantees it survives a compiler upgrade.

## The fix

**Library side** (trade-vex/disruptor-rs#3): handlers receive `&UnsafeCell<E>` — the one type for
which aliasing plus interior mutation is defined — and the disjointness contract is written down
where callers meet it. No storage change was needed; `RingBuffer` already stored `UnsafeCell<E>` and
was discarding the cell to produce a `*mut E`.

**Here**: establish a single owner *before* creating the mutable reference. Creating the reference is
the UB, so it is not enough to only *write* under a shard predicate — the predicate has to gate the
reference itself:

```rust
// SAFETY: This is a shared scalar read; peer handlers may read user_id
// concurrently, and no handler in this barrier group mutates user_id.
let user_id = unsafe { (*cell.get()).user_id };
if (user_id & shard_mask) != $shard_id as u64 { return; }
// SAFETY: Journaling is ordered before R1, and the shard predicate above is
// exclusive, so only this R1 handler creates a mutable reference for this sequence.
let cmd = unsafe { &mut *cell.get() };
```

| Handler group | Status | Why |
|---|---|---|
| R1 | sound | shard predicate on `user_id` gates the reference |
| Matching | sound | shard predicate on `market_id` gates the reference |
| Journaling | sound | sole handler in its group — and #158 orders it before R1 |
| Events | sound | sole handler in its group |
| **R2** | **not sound — see #183** | taker and maker may be on different shards |

**The dependency on #158 is load-bearing.** Without the `and_then()` it inserts, journaling shares a
barrier with R1 and holds a whole-command `&mut` concurrently with R1's owner — which would defeat
the R1 fix entirely. That is why this is stacked rather than cut from develop.

## What this does not fix

R2 is entered by all four shards and does work if it owns *either* the taker or any maker, so two
handlers legitimately settle one command concurrently. There is no single owner to gate on.

A projection fix was written and **reverted during review**: settling into a local
`MatcherTradeEvent` and copying `maker_balance` back changes behaviour, because `settle_trade` writes
that field for *both* participants — so the surviving value would depend on whether taker and maker
share a shard. That is new nondeterminism in settlement, from a change meant to be a pure soundness
fix. R2 therefore keeps its whole-command `&mut` and a comment that says plainly it is not sound,
rather than one asserting a safety it does not have.

Tracked in **#183**, which needs `handle_trade_event` to take projected fields — a `processors/`
change, and one that should resolve the `maker_balance` double-write first.

`cargo test --workspace`: 153 passed, 0 failed. clippy: 0 warnings. No existing test body changed.

## Related

Closes #128

- vex-core #183 — the R2 residue split out of this.
- vex-core #127 / PR #158 — the barrier this is stacked on.
- trade-vex/disruptor-rs#3 — the library-side change this depends on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of order processing across risk, matching, journaling, and event workflows.
  * Strengthened processing safeguards to prevent conflicting updates during concurrent operations.
  * Improved event sequencing to ensure dependent processing occurs in the correct order.
  * Updated test coverage and supporting infrastructure to reflect the improved processing behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,7 +558,7 @@ dependencies = [
 [[package]]
 name = "disruptor"
 version = "3.6.0"
-source = "git+https://github.com/trade-vex/disruptor-rs.git?rev=19ef2a9#19ef2a9a04de0fb97bc09d852ce5da98396efb57"
+source = "git+https://github.com/trade-vex/disruptor-rs.git?rev=7e07a8b0218d767d021925e58c5ff20fd5c30d70#7e07a8b0218d767d021925e58c5ff20fd5c30d70"
 dependencies = [
  "core_affinity",
  "crossbeam-utils",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ thiserror = "2.0.12"
 rusteron-client = "=0.1.157"
 rusteron-media-driver = "=0.1.157"
 rusteron-archive = "=0.1.157"
-disruptor = {git="https://github.com/trade-vex/disruptor-rs.git" , rev="19ef2a9"}
+disruptor = {git="https://github.com/trade-vex/disruptor-rs.git" , rev="7e07a8b0218d767d021925e58c5ff20fd5c30d70"}
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tracing = "0.1"
--- a/networking/tests/client_server_tests.rs
+++ b/networking/tests/client_server_tests.rs
@@ -109,14 +109,20 @@ fn test_client_server_communication() {
         )
         .pin_at_core(1)
         .handle_events_with({
-            move |cmd: &mut OrderCommand, _, _| {
+            move |cell: &std::cell::UnsafeCell<OrderCommand>, _, _| {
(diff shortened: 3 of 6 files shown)


rahulghangas:
@codex full review

